Issue #31, opened 2021-05-30
Move config handling to koanf


fzipi:
Viper has been a pain for some of the config loads that I've been trying to do.

Let's give [koanf](https://github.com/knadh/koanf) a try and see if we can do better.

fzipi:
Implemented in #32.

PR #32, opened 2021-06-06
feat(config): move from viper to kaonf for proper env vars parsing

fzipi:

Changed from viper to koanf for reading configuration. It gives more flexibility, and also works for env variables.


fzipi:
Fixes #31 .

PR #33, opened 2021-06-13
feat(cookies): add cookie support

fzipi:

This PR add the last remaining part for compatibility with `ftw`: cookies.

While they are not used actively in the `coreruleset` tests, they are part of `ftw`, so we needed to cover that.

Cookies needed to be integrated at a higher level than connections with the tested waf. They are part of the HTTP client, and need to be active while we run all tests. That way when a server sends any cookie back, the client can add the cookies to the request.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -1,0 +1,79 @@
+package http
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net"
+	"net/http/cookiejar"
+	"strings"
+	"time"
+
+	"github.com/rs/zerolog/log"
+	"golang.org/x/net/publicsuffix"
+)
+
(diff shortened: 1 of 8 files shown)


http/client.go
```diff
@@ -0,0 +1,79 @@
...
+
+	err := c.Transport.Request(&req)
+
+	if err != nil {
+		log.Error().Msgf("http/client: error sending request: %s\n", err.Error())
+	} else {
+		response, err = c.Transport.Response()
+		if err != nil {
+			log.Debug().Msgf("ftw/run: error receiving response: %s\n", err.Error())
+			// This error might be expected. Let's continue
```
jcchavezs, line 69: I am curious about this one. Why we expect an error here?

PR #51, opened 2022-03-19
Use the local test web server for all tests

theseion:
Fixed issues with false positive test results

### DIFF
--- a/runner/run.go
+++ b/runner/run.go
@@ -2,6 +2,7 @@ package runner
 
 import (
 	"errors"
+	"fmt"
 	"regexp"
 	"strconv"
 	"time"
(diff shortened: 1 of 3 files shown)


theseion:
Note: not all tests will succeed. Originally, some tests succeeded even though they shouldn't have. After fixing that, they now fail. That will be cleaned up with introduction of log marking.

runner/run.go
```diff
@@ -35,6 +36,7 @@ func Run(include string, exclude string, showTime bool, output bool, ftwtests []
 			// if we received a particular testid, skip until we find it
 			if needToSkipTest(include, exclude, t.TestTitle, tests.Meta.Enabled) {
 				addResultToStats(Skipped, t.TestTitle, &stats)
+				log.Info().Msgf("Skipping test %s", t.TestTitle)
```
fzipi, line 39: The problem here is that we cannot control if this is logged in quiet mode.

fzipi (commented):
Other than that, looks nice!

fzipi:
So we should merge this one despite tests failing?

theseion:
Depends. The other PR is based on this one, so it includes these changes too. I just wanted to separate the change. Tests in the other PR are all green.

runner/run.go
theseion: I've changed it to respect quiet mode. Better?

PR #55, opened 2022-04-10
Open log file only once

theseion:


runner/run.go
```diff
@@ -26,13 +27,29 @@ import (
...
+	}
+	var logFile *os.File
+	if logLines.FileName != "" {
+		var err error
+		logFile, err = os.Open(logLines.FileName)
+		if err != nil {
+			log.Panic().Caller().Err(err).Msg("failed to open file")
+		}
+		defer logFile.Close()
+	}
```
fzipi, line 42: This can easily be moved to a singleton pattern, right?

theseion: Yes. The only other object that uses FTWLogLines is the check object. And the only reason I need FTWLogLines is because I put the methods there (they seem to fit well in there). If you have an idea, go for it, I wasn't sure what the best way would be to proceed.

PR #61, opened 2022-08-07
When overriding dest_addr, also override Host header

dkegel-fastly:
This is my local fix for https://github.com/fzipi/go-ftw/issues/60.
Seems to work well enough here, though the problem running 920171-1 which prompted the nil check could use a look.


theseion:
Replaced by #63.

Thanks for the contribution. I've polished your PR with a test and some cleanup, and added you as co-author to the commit.

PR #63, opened 2022-08-18
feat: set "Host" header from dest_addr override

### LINKED ISSUE #60: Overriding dest_addr does not override Host header, leading to connection failures on some cloud WAFs
This makes it hard to test WAFs running on hosting providers that rely on the Host header.

Overriding Host in applyInputOverride where it overrides DestAddr seems to fix this, draft PR coming.

theseion:
- "Host" header should be overridden with value of `dest_addr` override
  to ensure that requests will be accepted by targets.
- Added test for the above change.
- Converted `config.FTWConfig.TestOverride.Input` from `map[string]string`
  to `test.Input` for improved validation and better parsing

Co-authored-by: @dkegel-fastly

Replaces #61 
Fixes #60

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,7 +15,7 @@ import (
 // or uses `.ftw.yaml` as default file
 func NewConfigFromFile(cfgFile string) error {
 	// kaonf merges by default but we never want to merge in this case
-	FTWConfig = nil
+	Reset()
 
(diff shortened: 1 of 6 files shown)


fzipi (approved):
I see what you did there. Thanks.

PR #68, opened 2022-09-02
fix: close previous connection before starting new one

anuraaga:
I randomly noticed this while debugging flakiness in my ftw tests. It didn't actually help 😂 But though it's still generally a good idea to close connections since otherwise I guess it is relying on GC.


fzipi (approved):
Of course, makes sense.

PR #69, opened 2022-09-02
feat: make request timeout configurable

anuraaga:
The 1s timeout is pretty long but currently my ftw tests don't pass with it - I need to debug whether this is an Envoy configuration issue, wasm performance issue, whether it affects real traffic, etc, and hopefully it won't be slow. But in the meantime, being able to run the test suite by configuring a higher timeout would be really great.

It looks like some files weren't gofmt'd and my IDE automatically does it, I hope that's OK

### DIFF
--- a/runner/run_test.go
+++ b/runner/run_test.go
@@ -1,19 +1,19 @@
 package runner
 
 import (
-    "fmt"
-    "net/http"
-    "net/http/httptest"
-    "os"
-    "testing"
-    "time"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"regexp"
+	"testing"
 
-    "github.com/fzipi/go-ftw/check"
-    "github.com/fzipi/go-ftw/config"
-    "github.com/fzipi/go-ftw/ftwhttp"
-    "github.com/fzipi/go-ftw/test"
+	"github.com/rs/zerolog/log"
 
-    "github.com/rs/zerolog/log"
+	"github.com/fzipi/go-ftw/check"
+	"github.com/fzipi/go-ftw/config"
+	"github.com/fzipi/go-ftw/ftwhttp"
+	"github.com/fzipi/go-ftw/test"
 )
 
 var yamlConfig = `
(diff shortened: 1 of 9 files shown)


fzipi (approved):
I like the general idea.

One thing that is bothering me now is that the `Run` is taking way too many parameters now, and probably changing it to a `RunOptions` functional style will be more readable and easy to update. What do you think?

If you want to push a new PR with that, I can merge this one in the meantime.

runner/run_test.go
```diff
@@ -410,37 +411,54 @@ func TestRun(t *testing.T) {
...
+		if res := Run([]test.FTWTest{ftwTest}, Config{
+			Include: regexp.MustCompile("008"),
+			Exclude: regexp.MustCompile("0*"),
+		}); res.Stats.TotalFailed() > 0 {
 			t.Error("Oops, test run failed!")
 		}
 	})
 
 	t.Run("exclude test 010", func(t *testing.T) {
-		if res := Run("*", "010", false, false, []test.FTWTest{ftwTest}); res.Stats.TotalFailed() > 0 {
```
anuraaga, line 437: There were some issues with the tests using `"*"`, which isn't a valid regexp so was hitting the `err` branch of the matchers. Changing the programmatic way of running to accept `*regexp.Regexp` directly allows users to see such errors in the IDE, and for go-ftw to not need to worry about problematic regexps from user

runner/run_test.go
```diff
@@ -472,11 +490,11 @@ func TestOverrideRun(t *testing.T) {
 	}
 	replaceDestinationInTest(&ftwTest, *fakeDestination)
 
-	t.Run("override and execute all", func(t *testing.T) {
```
anuraaga, line 475: There were some `t.Run` in tests with only one case which is redundant so removed them

anuraaga (commented):
Thanks @fzipi - added a Config struct

runner/run_test.go
fzipi: Excellent. I think I saw this behavior before, so a welcomed change.

runner/run_test.go
fzipi: Cool!

fzipi:
Looks good! The only thing that I see is that there is a lot of `Config`s going around now :)

If you feel good about it, let's merge. I need to review afterward how is this going to look when used programmatically.

fzipi:
Just to be in sync. From this [doc](https://golang.cafe/blog/golang-functional-options-pattern.html) I was thinking of option 3, and you implemented option 2.

anuraaga:
Thanks @fzipi - yeah I think this is good to merge. Currently have to build from my fork so will be glad to switch back 😃

I tend to have a preference for option 2 mostly since it gives easier discoverability to the options just IMO so went with it. Reworking to 3 would be fine though. Just some notes from wazero in case interesting.

https://github.com/tetratelabs/wazero/blob/main/RATIONALE.md#why-arent-configuration-assigned-with-option-types

fzipi:
No worries, for now it makes sense, and simplifies the work. Let's merge and see it in action.

PR #71, opened 2022-09-06
Reformat code

anuraaga:
While it'll be good to add lint, in the meantime it'd be good to get the files to reformat to reduce noise in PRs like #69. Though it looks like I already touched a lot of the files 😂


fzipi (approved):
LGTM

PR #72, opened 2022-09-08
feat: Use httpbin `/status` endpoint for markers

theseion:
The root URI is not mapped by `httpbin`, so the repsponse will contain a large amount of HTML for the 404 page, which ends up in the log. We now use the `/status` endpoint to reduce the data transferred in each response to a marker request and the amount of data in the log.


fzipi (approved):
Looks good!

PR #73, opened 2022-09-09
chore: make testing rule more generic for different waf implementations

### LINKED ISSUE #64: Include log marker name in example CRS test rule
I noticed that when looking for a log marker, it also checks for the marker name

https://github.com/fzipi/go-ftw/blob/f5f64a16b3d2bebea600ea070ffd5baa7f36213c/waflog/read.go#L114

The example rule for logging the marker doesn't output the name though

https://github.com/fzipi/go-ftw#how-log-parsing-works

Presumably there is some assumption that the log message automatically includes header names - but would the rule be more generic by including the name in it, e.g.

```
# Write the value from the X-CRS-Test header as a marker to the log
SecRule REQUEST_HEADERS:X-CRS-Test "@rx ^.*$" \
  "id:999999,\
  phase:1,\
  log,\
  msg:'X-CRS-Test %{MATCHED_VAR}',\
  pass,\
  t:none"
```

anuraaga:
Fixes #64

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ SecRule REQUEST_HEADERS:X-CRS-Test "@rx ^.*$" \
   "id:999999,\
   phase:1,\
   log,\
-  msg:'%{MATCHED_VAR}',\
+  msg:'X-CRS-Test %{MATCHED_VAR}',\
   pass,\
   t:none"
 ```


theseion:
Thanks!

PR #76, opened 2022-09-21
fix: remove RTT tracking from client/transport in favor of direct tra…

anuraaga:
…cking.

Currently in non-cloud mode, the RTT printed is always a few nanoseconds because the second marker request initializes a new connection, which resets the RTT struct. The collection of RTT within the client/connection doesn't seem to provide much right now - while it could possibly be refactored to actually do the start/stop tracking within `Do` or similar, it doesn't seem worth it and I've removed it in favor or just measuring at the caller.


theseion:
Great find! Thanks for the input.
I've quickly whipped up an alternative that doesn't reinitialize `Transport`. I'll open a PR and let @fzipi decide what he likes better.

fzipi:
Yeah, good find! The first versions used direct timing in the client but at some point I needed to pass that for something. Probably for testing? 🤔 Don't remember now.

Let's wait for @theseion alternative and decide what's best.

fzipi:
My impression is that #77 also optimizes the marker for reusing connections instead of recreating them.

But now my question is: what are timings in our context? What do we want to track? Moreover, if a test has many stages, the correct time is from start to finish in all the stages. So maybe moving the time up to the runner would make sense to simplify as this PR proposes.

I think that, conceptually, the whole time-tracking idea was originally made for the purpose of getting the correct subset of log lines. Now that having markers is a thing, there is a grey line on that particular usage, or to know exactly when a stage has been run. This reinforces the idea of not needing great accuracy (by getting timings down from the client, if there was a difference), and having a simpler implementation that gives us consistent numbers.

anuraaga:
For context, I use the timings to see the overhead added by enabling WAF through comparison. So enabling WAF or not for Envoy, setting FTW to cloud mode so it doesn't look for marker lines (not present in the WAF disabled case) and seeing how long it takes to handle each request - the number does not have to be super-precise and the current logic works fine for me.

fzipi:
Excellent, so let's merge this one.

fzipi:
I meant close, and I did merge 🤦 . Just reverted.

PR #77, opened 2022-09-22
fix: Reuse connection transport for marker requests

theseion:
Marker requests would reset the `Transport` struct, thereby unintentionally resetting the round trip time.

- Add `NewOrReusedConnection` to allow connections to be reused
- Move dialling logic to separate method
- Make marker requests use `NewOrReusedConnection` instead of `NewConnection`

This is an alternative to #76. @fzipi you need to decide which one you want to keep.

### DIFF
--- a/ftwhttp/client.go
+++ b/ftwhttp/client.go
@@ -36,36 +36,53 @@ func NewClient(config ClientConfig) *Client {
 
 // NewConnection creates a new Connection based on a Destination
 func (c *Client) NewConnection(d Destination) error {
-	if c.Transport != nil {
+	if c.Transport != nil && c.Transport.connection != nil {
 		if err := c.Transport.connection.Close(); err != nil {
 			return err
 		}
 	}
 
-	var err error
-	var netConn net.Conn
+	c.Transport = &Connection{
+		protocol:    d.Protocol,
+		readTimeout: c.config.ReadTimeout,
+		duration:    NewRoundTripTime(),
+	}
 
+	netConn, err := c.dial(d)
+	if err == nil {
+		c.Transport.connection = netConn
+	}
+
+	return err
+}
+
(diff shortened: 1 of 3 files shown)


anuraaga:
Thanks @theseion - my personal opinion is if keeping the RTT struct, we should go ahead and fix so it's not just a replacement for `time.Now()` at the caller as it currently is. My understanding is this is the main motivation for this PR.

If not doing that though, the simplification in #76 seems worth it - and we could still do this reuse but maybe it's not so meaningful then. I don't know if `Do` would ever have enough logic to make timings more precise by embedding them so lean towards simplifying by removing RTT from the client.

theseion:
I agree, currently, using explicit timing and using the RTT struct will report the exact same thing, and it's not actually the round trip time but the time between start and stop. In between those two statements a bunch of requests will have happened. So maybe we need to change the idea to "record average RTT of requests". The output should reflect that and maybe show both, e.g.:
```
test time: 23ms
avg rtt: 14ms
```

fzipi:
I think that even if the RTT timing is broken, this patch is needed to optimize reuse in connections. Is that right? In that case, we should merge this one to make the optimization available.

Then we can discuss if/how we want to track timings.

fzipi (approved):

anuraaga:
If reusing the connection works well then this does seem like a good idea. Just want to confirm though - I had thought arbitrarily that not reusing may have been intentional. We send a lot of weird requests, and it wouldn't surprise me if the connection gets corrupt in some cases but have no idea actually.

theseion:
I don't think reusing the connection is of much significance in terms of performance, since the actual `net.Conn` will be recreated anyway; but it's certainly nicer. However, now that I've looked at the code again, there's a leak when reusing the connection, because the old `net.Conn` is never closed. I'll open a new PR for that.

theseion:
Seee #84.

PR #79, opened 2022-09-22
Directly read HTTP response from connection instead of copying to byt…

anuraaga:
…es first.

Currently, there are cases where go-ftw will block until read timeout even after getting a whole HTTP response. I'm not exactly sure what causes it (I notice in my tests when Envoy returns an error response directly without proxying it seems to happen), but it sort of makes sense - usually you'd expect to need to parse the HTTP response to know when it's done since a connection is just otherwise a stream of bytes, so I'm actually less sure why it wouldn't always be blocking until read timeout.

This changes to parse straight into the HTTP response so HTTP semantics correctly let the response parsing finish when it's done, now I see some tests that always passed after 1s taking the expected several ms.

### DIFF
--- a/ftwhttp/connection.go
+++ b/ftwhttp/connection.go
@@ -3,6 +3,7 @@ package ftwhttp
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
 	"io"
 	"net"
@@ -104,13 +105,20 @@ func (c *Connection) Response() (*Response, error) {
 		return nil, err
 	}
 
-	reader := *bufio.NewReader(r)
+	buf := &bytes.Buffer{}
+
+	reader := *bufio.NewReader(io.TeeReader(r, buf))
 
 	httpResponse, err := http.ReadResponse(&reader, nil)
 	if err != nil {
 		return nil, err
 	}
+
+	data := buf.Bytes()
+	log.Trace().Msgf("ftw/http: received data - %q", data)
+
 	response := Response{
+		RAW:    data,
 		Parsed: *httpResponse,
 	}
 	return &response, err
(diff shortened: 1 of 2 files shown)


ftwhttp/connection.go
```diff
@@ -108,21 +98,19 @@ func (c *Connection) Request(request *Request) error {
...
 
-	r := bytes.NewReader(data)
 	reader := *bufio.NewReader(r)
 
 	httpResponse, err := http.ReadResponse(&reader, nil)
 	if err != nil {
 		return nil, err
 	}
 	response := Response{
-		RAW:    data,
```
anuraaga, line 125: This wasn't used.

ftwhttp/connection.go
```diff
@@ -65,25 +64,16 @@ func (c *Connection) send(data []byte) (int, error) {
...
-	}
-
-	if neterr, ok := err.(net.Error); ok && !neterr.Timeout() {
-		log.Error().Msgf("ftw/http: %s\n", err.Error())
-	} else {
-		err = nil
+	if err := c.connection.SetReadDeadline(time.Now().Add(c.readTimeout)); err != nil {
+		return nil, err
 	}
-	log.Trace().Msgf("ftw/http: received data - %q", buf)
```
anuraaga, line 84: Lost this, not sure of a good way to preserve it though. If anyone knows an alternative to this approach that still lets the response parsing finish without read timeout would be happy to change to it.

anuraaga: Was able to use TeeReader

fzipi: Excellent! TIL.

PR #80, opened 2022-09-23
feat: testoverride ignore/forcefail/forcepass as regexes

M4tteoP:
In this PR:
- `ignore`, `forcefail` and `forcepass` lists are now considered regexes.
- Updated example inside the readme accordingly

Closes https://github.com/fzipi/go-ftw/issues/75.


fzipi:
I guess you should rebase after moving to a new organization.

config/types.go
```diff
@@ -37,3 +42,10 @@ type FTWTestOverride struct {
 	ForcePass map[string]string `koanf:"forcepass"`
 	ForceFail map[string]string `koanf:"forcefail"`
 }
+
+// FTWTestOverrideRe hold the lists transformed into regexes
+type FTWTestOverrideRe struct {
```
anuraaga, line 47: koanf seems to support `TextUnmarshaler` so I believe you can avoid needing a second field / struct which makes the configuration a bit harder to reason about

Something like

```go

type FTWTestOverride struct {
  Ignore map[string]FTWRegexp
}

type FTWRegexp regexp.Regexp

func (r *FTWRegexp) UnmarshalText(b []byte) error {
  re, err := regexp.Compile(string(b))
  if err != nil {
    return fmt.Errorf("invalid regexp: %w", err)
  }
  *r = re
  return nil
}
```

config/types.go
```diff
@@ -21,8 +25,9 @@ var FTWConfig *FTWConfiguration
 type FTWConfiguration struct {
 	LogFile             string          `koanf:"logfile"`
 	TestOverride        FTWTestOverride `koanf:"testoverride"`
-	LogMarkerHeaderName string          `koanf:"logmarkerheadername"`
```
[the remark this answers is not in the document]
M4tteoP: Sorry, I just added `TestOverrideRe` in the middle of the struct, so `LogMarkerHeaderName` appears to be removed, but actually is right after `TestOverrideRe`. If I manage to implement anuraaga suggestion, this confusing line will be removed.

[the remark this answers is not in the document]
theseion: You're right. Sorry :)

config/types.go
M4tteoP: Implemented, thank you! I had to update Koanf and the resulting map is `map[*FTWRegexp]string`. If we want to keep the excluded id - comment map, the key was the id (now regex) and the string value is the comment.

README.md
```diff
@@ -184,7 +184,8 @@ Tests can be altered using four lists:
 - `forcepass` is for tests you want to pass unconditionally. You should add a comment on why you force to pass the test
 - `forcefail` is for tests you want to fail unconditionally. You should add a comment on why you force to fail the test
 
-Example using all the lists above:
+The match is performed using Go regexp. You can whether match a specific test or all of them about a specific rule, or file.
```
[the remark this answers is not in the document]
M4tteoP: Hi! Thanks, I updated the Readme. I think that they are still lists. Each element of the list permits to have 1-n matches against test IDs, but for each ignore/forcefail/forcepass we can have more than one regex. E.g:
Eg:
```
  forcepass:
    '123456-02$': 'This test will always pass'
    '123457-.*': 'All the tests about rule 123457 will always pass' 
```
Waiting for your feedback about the updated proposal :) Thank you!

config/types.go
theseion: Oh. Yes, you're absolutely right. I don't know what I was thinking...

README.md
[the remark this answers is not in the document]
theseion: Oh. Yes, you're absolutely right. I don't know what I was thinking...

check/base.go
```diff
@@ -60,20 +62,32 @@ func (c *FTWCheck) SetNoLogContains(contains string) {
 
 // ForcedIgnore check if this id need to be ignored from results
 func (c *FTWCheck) ForcedIgnore(id string) bool {
-	_, ok := c.overrides.Ignore[id]
-	return ok
+	for re, _ := range c.overrides.Ignore {
+		if (*regexp.Regexp)(re).MatchString(id) {
```
anuraaga, line 66: Let's go ahead and add matchstring to the type itself

```
func (r *FTWRegexp) MatchString(s string) bool {
	return (*regexp.Regexp)(r).MatchString(s)
}
```

config/config.go
```diff
@@ -69,7 +72,7 @@ func NewConfigFromEnv() error {
 
 // NewConfigFromString initializes the configuration from a yaml formatted string. Useful for testing.
 func NewConfigFromString(conf string) error {
-	// kaonf merges by default but we never want to merge in this case
+	// kaonf merges by default, but we never want to merge in this case
```
theseion, line 75:
```suggestion
	// koanf merges by default, but we never want to merge in this case
```

config/config.go
```diff
@@ -39,14 +39,17 @@ func NewConfigFromFile(cfgFile string) error {
...
+	}
 	loadDefaults()
 
 	return err
 }
 
 // NewConfigFromEnv reads configuration information from environment variables that start with `FTW_`
 func NewConfigFromEnv() error {
-	// kaonf merges by default but we never want to merge in this case
+	// kaonf merges by default, but we never want to merge in this case
```
theseion, line 52:
```suggestion
	// koanf merges by default, but we never want to merge in this case
```

config/config.go
```diff
@@ -14,7 +14,7 @@ import (
 // NewConfigFromFile reads configuration information from the config file if it exists,
 // or uses `.ftw.yaml` as default file
 func NewConfigFromFile(cfgFile string) error {
-	// kaonf merges by default but we never want to merge in this case
+	// kaonf merges by default, but we never want to merge in this case
```
theseion, line 17:
```suggestion
	// koanf merges by default, but we never want to merge in this case
```

config/config_test.go
```diff
@@ -34,111 +35,112 @@ var jsonConfig = `
...
-	assert.NoError(t, err)
+    err := NewConfigFromString(yamlConfig)
+    assert.NoError(t, err)
 }
 
 func TestNewConfigFromEnv(t *testing.T) {
-	// Set some environment so it gets merged with conf
-	os.Setenv("FTW_LOGFILE", "kaonf")
+    // Set some environment so it gets merged with conf
+    os.Setenv("FTW_LOGFILE", "kaonf")
```
theseion, line 95:
```suggestion
    os.Setenv("FTW_LOGFILE", "koanf")
```

config/config_test.go
```diff
@@ -34,111 +35,112 @@ var jsonConfig = `
...
 
-	err := NewConfigFromEnv()
-	assert.NoError(t, err)
+    err := NewConfigFromEnv()
+    assert.NoError(t, err)
 
-	assert.Equal(t, "kaonf", FTWConfig.LogFile)
+    assert.Equal(t, "kaonf", FTWConfig.LogFile)
```
theseion, line 100:
```suggestion
    assert.Equal(t, "koanf", FTWConfig.LogFile)
```

config/types.go
```diff
@@ -32,8 +37,24 @@ type FTWConfiguration struct {
...
+	Input     test.Input            `koanf:"input"`
+	Ignore    map[*FTWRegexp]string `koanf:"ignore"`
+	ForcePass map[*FTWRegexp]string `koanf:"forcepass"`
+	ForceFail map[*FTWRegexp]string `koanf:"forcefail"`
+}
+
+type FTWRegexp regexp.Regexp
+
+func (r *FTWRegexp) UnmarshalText(b []byte) error {
+	fmt.Println("Unmarshalling!")
```
theseion, line 49: Use the logger. I don't think we want to see this message all the time. I also think the text could be more helpful. Maybe just remove the line entirely.

theseion (changes requested):

config/types.go
M4tteoP: Thanks, my bad, just a debugging print, not intended to be here.

theseion (approved):

anuraaga (approved):

PR #88, opened 2022-10-04
feat: publish docker images to ghcr with release

anuraaga:
It would be very convenient to have docker images for go-ftw. ghcr requires no set up or credential sharing and is a great option for OSS projects so recommend trying it out


.goreleaser.yml
```diff
@@ -16,6 +16,7 @@ builds:
       - darwin
     goarch:
       - amd64
+      - arm64
```
anuraaga, line 19: Should build this for M1 macs

anuraaga:
Looks like this: https://github.com/anuraaga/go-ftw/pkgs/container/go-ftw

fzipi:
Seems reasonable. Can we fix EOL at EOF?

fzipi:
Shall we also update the readme with this option?

anuraaga:
> Seems reasonable. Can we fix EOL at EOF?

Sure fixed

> Shall we also update the readme with this option?

Probably good to do this after a release (knowing that it's easy to forget though 😅 )

fzipi:
@anuraaga Do you mind using conventional commits syntax in PRs?

anuraaga:
@fzipi Sure updated

.github/workflows/release.yml
```diff
@@ -19,6 +19,13 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: 1.16
```
fzipi, line 21: Do you want to update go version also?

Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM gcr.io/distroless/static-debian11:debug
```
fzipi, line 1: Is there a reason for using debug? Maybe for initial testing and then we remove it?

anuraaga: Actually sorry, take that back. I think it will be common to use this as a base image, to embed scripts, or otherwise allow control of options with environment variables. The non-debug image doesn't allow this as it has no shell at all. The only difference is the presence of busybox - what do you think?

fzipi: If that's the case, why don't we use alpine for example? I see the case for distroless for a service, but not for ftw...

fzipi (approved):
LTGM! 🚀

PR #91, opened 2022-10-07
feat: use goreleaser default naming for published archives

anuraaga:
This is just style so no worries if not going with this. I want to propose using goreleaser default naming scheme for release archives. goreleaser is very common now, and I like the consistency of archive names I see across various projects when writing tooling. I think it's worth doing that here.

Also publishes a windows-amd64 build (not arm64 since there is no viable platform for that yet)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,20 +11,16 @@ builds:
       - GO111MODULE=on
       - CGO_ENABLED=0
     binary: ftw
-    goos:
-      - linux
-      - darwin
-    goarch:
-      - amd64
-      - arm64
+    targets:
+      - linux_amd64
+      - linux_arm64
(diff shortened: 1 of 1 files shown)


jcchavezs (approved):

fzipi:
Does the naming change? If it does, can you provide also a PR for https://github.com/coreruleset/coreruleset/blob/cf383e1625c973bdc673307790be99ecbcbe8314/.github/workflows/go-ftw.yml#L28 so we can merge when releasing?

fzipi:
@anuraaga ☝️ ?

anuraaga:
Sorry for delay @fzipi sent https://github.com/coreruleset/coreruleset/pull/2846

PR #92, opened 2022-10-07
fix: simplify sonar workflow and only run on push

anuraaga:
The sonar workflow has always been failing because it attempts to download an artifact uploaded from a different workflow - artifacts are scoped to builds so different workflows can't access them like that. From what I can tell, processing coverage with sonar is to show coverage report in sonarqube. Instead we should upload to codecov so it's open, and I can add that separately if this makes sense.

I also removed running Sonar on pull request because it won't work for PRs from forked repos due to the use of a secret - generally we can't run checks that require secrets on PRs in OSS projects so best to just fix forward if main fails.

### DIFF
--- a/.github/workflows/sonar.yaml
+++ b/.github/workflows/sonar.yaml
@@ -1,33 +1,19 @@
 name: Sonar
 on:
-  workflow_run:
-    workflows: [test]
-    types: [completed]
+  push:
+    branches: [ main ]
+
 jobs:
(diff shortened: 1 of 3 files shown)


jcchavezs:
SGTM unless @fzipi has a concern on using codecov related to org restrictions or rules.

fzipi (approved):
LGTM.

PR #96, opened 2022-10-30
fix: properly URL encode data for POST requests

theseion:
While doing tests using unencoded data, we figured out that some parts of the encoding were not working.

This PR fixes this problem.


fzipi (approved):
This looks awesome.

PR #99, opened 2022-10-31
fix: remove most uses of log.Fatal when running programatically

anuraaga:
go-ftw provides a programmatic entry point for tests like this one

https://github.com/corazawaf/coraza/blob/v3/dev/testing/coreruleset/coreruleset_test.go#L213

There are some quirks like global config (will look at in another PR) and the code causing the process to exit because of use of `log.Fatal`, making test failures difficult to understand. This replaces `log.Fatal` with propagating errors. There is one remaining Fatal for a regexp which will require some less mechanical change but that also should have lower chance of triggering

### DIFF
--- a/runner/run_test.go
+++ b/runner/run_test.go
@@ -417,49 +417,55 @@ func TestRun(t *testing.T) {
 	replaceDestinationInTest(&ftwTest, *dest)
 
 	t.Run("show time and execute all", func(t *testing.T) {
-		res := Run([]test.FTWTest{ftwTest}, Config{
+		res, err := Run([]test.FTWTest{ftwTest}, Config{
 			ShowTime: true,
 			Quiet:    true,
 		})
+		assert.NoError(t, err)
 		assert.Equalf(t, res.Stats.TotalFailed(), 0, "Oops, %d tests failed to run!", res.Stats.TotalFailed())
 	})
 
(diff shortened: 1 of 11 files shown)


jcchavezs (approved):

runner/run_test.go
```diff
@@ -547,16 +555,17 @@ func TestBrokenPortOverrideRun(t *testing.T) {
...
 	})
+	assert.NoError(t, err)
 	assert.LessOrEqual(t, 0, res.Stats.TotalFailed(), "Oops, test run failed!")
 }
 
 func TestDisabledRun(t *testing.T) {
 	t.Cleanup(config.Reset)
 
-	err := config.NewConfigFromString(yamlConfig)
+	err := config.NewConfigFromString(yamlCloudConfig)
```
theseion, line 568: Why did you have to change the config here?

anuraaga: Thanks for checking forgot to mention it previously the presence of a log file was checked lazily so if we the test didn't run, like this one checking behavior of disabled tests, it was ok without a log file set. Now that validation is eager, which should generally be good. It seemed simplest to use cloud mode here which doesn't need a log file instead of setting up an unused file

theseion (approved):
Awesome! Thanks!

PR #102, opened 2022-11-04
fix(check): explain why files are not properly formatted

fzipi:

- when `check` fails, it does it more decently now
- shows the error and a possible fix


test/files.go
```diff
@@ -68,3 +71,24 @@ func readFileContents(fileName string) (contents []byte, err error) {
...
+		return "This might refer to a \"status\" line being '200', where it should be '[200]'. " +
+			"The default \"status\" is a list now."
+	}
+	matched, err = regexp.MatchString(`.*cannot unmarshal \[]interface {} into Go struct field FTWTest.Tests of type string.*`, yamlError.Error())
+	if err != nil {
+		return err.Error()
+	}
+	if matched {
+		return "This might refer to \"data\" on the test being a list of strings instead of a proper YAML multiline. " +
+			"To fix this, convert this \"data\" string list to a multiline YAML and this will be fixed. "
```
theseion, line 90: How would one do that? With `|`? Add a short example.

theseion (changes requested):

test/files.go
fzipi: Added!

theseion (approved):
Nice.

PR #103, opened 2022-11-11
fix: only escape key/values for data in POST requests.

anuraaga:
When trying to update FTW, I noticed CRS tests failing and AFAICT, it is because of the escaping behavior introduced in #96. Now, there are never key/value pairs since equals and ampersands are always escaped. Ideally, the test format itself was a yaml map instead of a single string or the single string was always escaped, but as is now, I think we have to assume that `=` and `&` in the `data` field are the form parameter delimiters and only escape the actual key/values.

### DIFF
--- a/ftwhttp/request_test.go
+++ b/ftwhttp/request_test.go
@@ -245,6 +245,10 @@ func TestRequestEncodesPostData(t *testing.T) {
 			encoded: "foo+bar",
 		},
 		{
+			raw:     "name=panda&food=bamboo",
+			encoded: "name=panda&food=bamboo",
+		},
+		{
 			// Test adding semicolons to test parse
 			raw:     `c4= ;c3=t;c2=a;c1=c;a1=/;a2=e;a3=t;a4=c;a5=/;a6=p;a7=a;a8=s;a9=s;a10=w;a11=d;$c1$c2$c3$c4$a1$a2$a3$a4$a5$a6$a7$a8$a9$a10$a11`,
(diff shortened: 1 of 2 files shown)


ftwhttp/request_test.go
```diff
@@ -228,23 +228,55 @@ func TestRequestURLParseFail(t *testing.T) {
...
+		},
+		{
+			// Test adding semicolons to test parse
+			raw:     `c4= ;c3=t;c2=a;c1=c;a1=/;a2=e;a3=t;a4=c;a5=/;a6=p;a7=a;a8=s;a9=s;a10=w;a11=d;$c1$c2$c3$c4$a1$a2$a3$a4$a5$a6$a7$a8$a9$a10$a11`,
+			encoded: "c4=+%3Bc3%3Dt%3Bc2%3Da%3Bc1%3Dc%3Ba1%3D%2F%3Ba2%3De%3Ba3%3Dt%3Ba4%3Dc%3Ba5%3D%2F%3Ba6%3Dp%3Ba7%3Da%3Ba8%3Ds%3Ba9%3Ds%3Ba10%3Dw%3Ba11%3Dd%3B%24c1%24c2%24c3%24c4%24a1%24a2%24a3%24a4%24a5%24a6%24a7%24a8%24a9%24a10%24a11",
+		},
+		{
+			// Already encoded
+			raw:     "foo+bar",
+			encoded: "foo+bar",
```
theseion, line 255: Add another that includes ampersand.

theseion (changes requested):
Thanks for the effort. We need to clean this up properly at some point. For example, I think we need to be able to control:
- space endoding (`%20` or `+`)
- parameter encoding (`=`, `&`) since the current encoding format is just a convention

We should also properly separate encoding of the query string and post body data.

fzipi (approved):

theseion (approved):

jcchavezs (approved):

PR #107, opened 2022-11-14
fix: look more than one line for log markers

### LINKED ISSUE #104: Error: If apache runs on loglevel INFO and write 404 into error.log, go-ftw can't find marker.
If `/status/200` is not existing and if apache is set to loglevel `info`, we get the following error message(s) in the log:

```
[2022-11-12 23:08:18.012572] [-:error] 127.0.0.1:36126 Y3AZUo3Gja4gB-tPE9uasgAAAA4 [client 127.0.0.1] ModSecurity: Warning. Unconditional match in SecAction. [file "/apache/conf/httpd.conf_pod_2022-11-12_22:23"] [line "265"] [id "999999"] [msg "x-crs-test 6c7427f7-6638-46fd-8ad5-aa66961d3ffe"] [hostname "localhost"] [uri "/status/200"] [unique_id "Y3AZUo3Gja4gB-tPE9uasgAAAA4"]
[2022-11-12 23:08:18.013007] [core:info] 127.0.0.1:36126 Y3AZUo3Gja4gB-tPE9uasgAAAA4 AH00128: File does not exist: /apache/htdocs/status/200
```

The 404 message prevents go-ftw from identifying and parsing the marker line correctly.

theseion:
Look back an additional line and do not assume that the marker will appear on the last line. httpd will write the 404 error **after** the ModSecurity output (depending on the log level).

This commit also fixes some casing issues that might occur, especially in case of tests.

Fixes #104.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,8 +112,8 @@ logfile: 'tests/logs/modsec3-nginx/error.log'
 
 This is the help for the `run` command:
 ```bash
-❯ ./ftw run --help
-Run all tests below a certain subdirectory. The command will search all y[a]ml files recursively and pass it to the test engine.
+./ftw run --help
+ Run all tests below a certain subdirectory. The command will search all y[a]ml files recursively and pass it to the test engine.
 
 Usage:
   ftw run [flags]
@@ -126,7 +126,10 @@ Flags:
   -h, --help                       help for run
       --id string                  (deprecated). Use --include matching your test only.
   -i, --include string             include only tests matching this Go regexp (e.g. to include only tests beginning with "91", use "91.*").
-      --output-failures-only       output only the result of failed tests
+      --max-marker-log-lines int   maximum number of lines to search for a marker before aborting (default 500)
+      --max-marker-retries int     maximum number of times the search for log markers will be repeated.
+                                   Each time an additional request is sent to the web server, eventually forcing the log to be flushed (default 20)
+      --output-failures-only       output only the results of failed tests
   -q, --quiet                      do not show test by test, only results
       --read-timeout duration      timeout for receiving responses during test execution (default 1s)
   -t, --time                       show time spent per test
@@ -136,7 +139,6 @@ Global Flags:
       --config string   override config file (default is $PWD/.ftw.yaml)
       --debug           debug output
       --trace           trace output: really, really verbose
-
 ```
 
 Here's an example on how to run your tests recursively in the folder `tests`:
--- a/waflog/read.go
+++ b/waflog/read.go
@@ -80,16 +80,19 @@ func (ll *FTWLogLines) getMarkedLines() [][]byte {
 }
 
 // CheckLogForMarker reads the log file and searches for a marker line.
-// logFile is the file to search
 // stageID is the ID of the current stage, which is part of the marker line
-func (ll *FTWLogLines) CheckLogForMarker(stageID string) []byte {
+// readLimit is the maximum numbers of lines to check
+//
+// If `readLimit` is reached while searching for a marker, the method will return
+// `true`, to signal that it aborted.
+func (ll *FTWLogLines) CheckLogForMarker(stageID string, readLimit int) ([]byte, bool) {
 	offset, err := ll.logFile.Seek(0, os.SEEK_END)
 	if err != nil {
 		log.Error().Caller().Err(err).Msgf("failed to seek end of log file")
-		return nil
+		return nil, false
 	}
 
-	// Lines in modsec logging can be quite large
+	// Lines in logging can be quite large
 	backscannerOptions := &backscanner.Options{
 		ChunkSize: 4096,
 	}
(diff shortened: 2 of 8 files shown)


waflog/read.go
```diff
@@ -102,15 +102,22 @@ func (ll *FTWLogLines) CheckLogForMarker(stageID string) []byte {
...
 		line, _, err = scanner.LineBytes()
 	}
-	if err != nil {
-		if err == io.EOF {
-			return nil
+
+	// Try to find the marker in the last and the next to last
+	// line. Some errors in httpd will be printed **after** the
+	// ModSecurity output (e.g. 404 with filesystem lookup).
+	for i := 1; i >= 0; i-- {
```
anuraaga, line 109: Can this be made configurable? For FTW tests with coraza-proxy-wasm, the logs are combined with Envoy logs - we [grep](https://github.com/corazawaf/coraza-proxy-wasm/blob/main/ftw/docker-compose.yml#L40) into a separate lot for FTW to separate them, but if the lookback could be configured to a high-ish number in those tests, we could probably get rid of that

dune73:
This is probably working in this case, but more verbose Apache will lead to more output in the error.log. So it feels a bit like a local workaround and not like a general solution.

I mean how expensive is walking backwards in the log if the last line does not contain the marker? I see the performance impact, but usually it will be the marker, so no problem...

Conceptually it should work with `LogLevel trace8` IMHO.

fzipi (commented):
I would say keep going back until you find the marker, or we hit EOF.

dune73:
If the file is huge, then going back until EOF could take forever and you repeat it 20 times ... I think there ought to be a reasonable counter.

Idea: 
* 1st status marker sent: Start on last line and go back 0 lines
* 2nd status marker sent: Start on last line and go back 1 lines
* 3rd status marker sent: Start on last line and go back 2 lines
* ...
* 20th status marker sent: Start on last line and go back 19 lines

waflog/read.go
theseion: > Why not just search until EOF?

Because you do not want to look through 200MB 20 times :)

Yes, we can absolutely make the lookback configurable. But maybe we can use a different approach. If we knew, which log lines we could ignore, we wouldn't have to pick an arbitrary number. So how about we add a configuration option for a regular expression (or multiple) that can filter unwanted log lines? In your case, you'll already know the pattern.

theseion:
> 1st status marker sent: Start on last line and go back 0 lines
> 2nd status marker sent: Start on last line and go back 1 lines
> 3rd status marker sent: Start on last line and go back 2 lines
> ...
> 20th status marker sent: Start on last line and go back 19 lines

That's not a bad idea. What do you think of [my other idea](https://github.com/coreruleset/go-ftw/pull/107#discussion_r1021955850)? Not sure exactly, how we would apply it generically though...

fzipi:
Nice idea, but we are leaving out Coraza and other implementations.

waflog/read.go
anuraaga: Thanks - I think a regex could work well for us. Though maybe we don't need it - is there any reason not to search until EOF or `X-CRS-Test`? It shouldn't result in ever going through an entire huge log file, and either the matched line will be for the current test, or the previous test and we'll retry and look again.

theseion:
Thanks for the feedback everyone. In conclusion, here's what I think we should do:
1. look back until EOF for the header (`X-CRS-Test`)
2. Stop if we find the header
3. Check the UUID against the line

Looking for the header without requiring the UUID to also match will prevent the search from reaching EOF every time and also means we don't need to use an arbitrary number to look back.

In addition, I can implement simple logic to stop the search early, for example, if we've reached EOF and read more than 100 lines without finding the header. Then it's likely that the header isn't in the file.

dune73:
That looks reasonable and the counter prevents you from stumbling over edge cases.

waflog/read.go
```diff
@@ -98,20 +101,33 @@ func (ll *FTWLogLines) CheckLogForMarker(stageID string) []byte {
...
 		line, _, err = scanner.LineBytes()
-	}
-	if err != nil {
-		if err == io.EOF {
-			return nil
+		if err != nil {
+			if err == io.EOF {
+				return nil, false
+			}
+			log.Trace().Err(err).Msg("found EOF while looking for log marker")
```
anuraaga, line 117: This line doesn't seem to be about EOF since that was handled above

anuraaga (approved):
Thanks, this is going to be a great UX improvement!

fzipi:
Reading the code again, I got to say this: I don't really like when we just add constants without clear meaning (e.g. 20 instead of `defaultMaxRetries` or similar), but mostly, not giving the user the power to decide how times to do something. I know we are using sane defaults, but other people's use cases might be solved by just using `--max-retries 100`, or `--read-max-bytes 1024`.

theseion:
+1000

waflog/read.go
theseion: Thanks. That was some copy pasta.

theseion:
I've opened a new issue to fix config default initialization: #108.

README.md
```diff
@@ -126,7 +126,10 @@ Flags:
   -h, --help                       help for run
       --id string                  (deprecated). Use --include matching your test only.
   -i, --include string             include only tests matching this Go regexp (e.g. to include only tests beginning with "91", use "91.*").
-      --output-failures-only       output only the result of failed tests
+      --max-marker-log-lines int   maximum number of lines to search for a marker before aborting (default 500)
+      --max-marker-retries int     maximum number of times the search for log markers will be repeated.
+                                   Each time an additional request is sent to the web server, eventually forcing the log to be flushed (default 20)
```
fzipi, line 131: ❤️

fzipi (dismissed):

fzipi (approved):
LGTM!

PR #110, opened 2022-11-17
Update develop with main changes

fzipi:
These changes slipped under the radar.


jcchavezs:
Does it make sense to keep this branching main/develop? Can't we just use main a do releases?

fzipi:
Probably it makes sense to use main only. But will double check things first.

PR #116, opened 2022-11-21
fix: remove goccy/go-yaml

### LINKED ISSUE #115: Reading all test files at startup takes several seconds
The function `GetTestsFromFiles` inf `files.go` takes several seconds to read all YAML test files into memory. While the number of tests has grown in recent weeks, the slowdown still feels significant.

anuraaga:
Fixes #115 

I could confirm this fixes the perf regression.

I suspect `describeYamlError` won't work after this, @fzipi could you follow up after this PR to fix it and add some unit tests for the cases?


fzipi:
My proposal is to downgrade the library instead of removing it. There are still some features we want, and in the future, it is the only one that brings additional YAML features like anchors and aliases that might help when writing tests.

So I'm proposing to close this one and go for #117.

fzipi (commented):
My proposal is to downgrade the library instead of removing it. There are still some features we want, and in the future, it is the only one that brings additional YAML features like anchors and aliases that might help when writing tests.

So I'm proposing to close this one and go for #117.

anuraaga:
For the record, which features are you thinking of? Generally it's good to reduce dependencies for software chain reasons so it should be something significant IMO

fzipi:
> There are still some features we want, and in the future, it is the only one that brings additional YAML features like anchors and aliases that might help when writing tests.

@anuraaga ☝️

anuraaga:
Sorry read those as future not now, oops.

As far as I can tell, those features are only for marshaling while go-yaml supports them fine for unmarshal. Does that mean it's not meaningful here?

fzipi:
Per https://github.com/goccy/go-yaml#2-reference-elements-declared-in-another-file, looks like I want that.

anuraaga:
IIUC it's the ability to define anchors in separate files instead of just one. If that's the case, I guess it's not actually yaml.

While go-ftw may be a primary implementer of the tests, it can't be the only one. I don't think it's safe for CRS to be defining tests in a way that does not support standard yaml parsers. There's already support for generating regexes so wouldn't we want to add tooling for generating compliant yaml instead of forcing parser implementations?

jcchavezs:
+1 on using standard yaml only.

fzipi:
I don't think this is the place to discuss this. If we want to, let's raise a new issue and discuss there.

fzipi:
So, in summary:
- there is no need to remove this library, as of now. The functionality we use out of the box is going to break when removed, so more work is needed to replace and/or re-implement it.
- while the desire is to keep the number of libraries used to a minimum (reusing vs. implementing something) when there is a library that provides the same feature, I don't think it is worth it for this kind of project. I can understand it for Coraza or a project that needs to be extremely performant. This is not the case.
- anchors and aliases are standard yaml: https://yaml.org/spec/1.2.2/#anchors-and-aliases. The inability of the go "standard" library to parse those are a problem for that library.

fzipi:
Also, the "standard" library gopkg.in/yaml.v3 is not part of the core, is just another library 😄 .

theseion:
I agree with @fzipi. Downgrade the library. They will notice the performance impact at some point and then we can upgrade.

PR #120, opened 2022-11-25
fix: create empty lists in stats

fzipi:

After running a couple of plugin tests, I realized that if some error was badly propagated, stats were created with `nil`, but at the `os.Exit` we were calling the stats, hence getting a fault.

Now this PR makes two changes:
- create Stats as empty lists instead of nil
- propagate errors so it can be shown at the cmd level

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -7,7 +7,6 @@ import (
 	"regexp"
 	"time"
 
-	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 
 	"github.com/coreruleset/go-ftw/output"
@@ -49,7 +48,6 @@ var runCmd = &cobra.Command{
 		tests, err := test.GetTestsFromFiles(files)
 
 		if err != nil {
-			log.Fatal().Err(err)
 			return err
 		}
 
@@ -78,7 +76,6 @@ var runCmd = &cobra.Command{
 		}, out)
 
 		if err != nil {
-			log.Fatal().Err(err)
 			return err
 		}
 		if currentRun.Stats.TotalFailed() > 0 {
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"regexp"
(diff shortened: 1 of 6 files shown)


cmd/run.go
```diff
@@ -71,11 +74,16 @@ var runCmd = &cobra.Command{
...
 		}, out)
+
 		if err != nil {
 			log.Fatal().Err(err)
+			return err
 		}
-
-		os.Exit(currentRun.Stats.TotalFailed())
+		if currentRun.Stats.TotalFailed() > 0 {
+			errText := fmt.Errorf("failed %d tests", currentRun.Stats.TotalFailed())
```
theseion, line 83: `Errorf` returns `error`, no need for `errors.New(...)`.

theseion (changes requested):

cmd/run.go
fzipi: Fixed.

fzipi:
My local golanci-lint pushed me to piggyback more changes unrelated to this one, about problems on output:
```
output/output.go:50:2: field `cat` is unused (unused)
	cat        catalog
	^
output/output.go:124:3: ineffectual assignment to text (ineffassign)
		text, found = normalCatalog[key]
```
So I've made use of those explicitly and wrote additional tests.

cmd/run.go
```diff
@@ -36,16 +38,19 @@ var runCmd = &cobra.Command{
...
-			log.Fatal().Msgf("You need to choose one: use --include (%s) or --exclude (%s)", include, exclude)
+			cmd.SilenceUsage = false
+			return fmt.Errorf("You need to choose one: use --include (%s) or --exclude (%s)", include, exclude)
 		}
 		files := fmt.Sprintf("%s/**/*.yaml", dir)
 		tests, err := test.GetTestsFromFiles(files)
 
 		if err != nil {
 			log.Fatal().Err(err)
+			return err
```
theseion, line 53: This will never be reached, will it? `Fatal` calls `Exit`.

cmd/run.go
```diff
@@ -71,11 +76,15 @@ var runCmd = &cobra.Command{
 			MaxMarkerRetries:  maxMarkerRetries,
 			MaxMarkerLogLines: maxMarkerLogLines,
 		}, out)
+
 		if err != nil {
 			log.Fatal().Err(err)
+			return err
```
theseion, line 82: Same here: `Fatal` calls `Exit`

theseion (changes requested):

cmd/run.go
fzipi: Of course, propagation is the key here.

cmd/run.go
fzipi: Same here: error should be propagated.

theseion (approved):

PR #124, opened 2022-12-02
feat(config): add default max marker as default values

### LINKED ISSUE #108: Implement proper config initialization for default values
For values that need a default other than the zero value (e.g., `20`), the configuration currently isn't initialized properly in tests, as the defaults would be set by cobra. Instead of adding the default values to each configuration that is being created in the tests, we should have a means to use a default configuration, e.g., by having a function `NewConfig`.

fzipi:

Fixes #108.

### DIFF
--- a/config/types.go
+++ b/config/types.go
@@ -25,7 +25,7 @@ const (
 
 // FTWConfiguration FTW global Configuration
 type FTWConfiguration struct {
-	// Logfile is the path to the file that contains the WAF logs to check
+	// Logfile is the path to the file that contains the WAF logs to check. The path may be absolute or relative, in which case it will be interpreted as relative to the current working directory.
 	LogFile string `koanf:"logfile"`
 	// TestOverride holds the test overrides that will apply globally
 	TestOverride FTWTestOverride `koanf:"testoverride"`
--- a/config/types.go
+++ b/config/types.go
@@ -17,14 +17,26 @@ const (
 	DefaultRunMode RunMode = "default"
 	// DefaultLogMarkerHeaderName is the default log marker header name
 	DefaultLogMarkerHeaderName string = "X-CRS-Test"
+	// DefaultMaxMarkerRetries is the default amount of retries that will be attempted to find the log markers
+	DefaultMaxMarkerRetries int = 20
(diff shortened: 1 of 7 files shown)


config/types.go
```diff
@@ -17,14 +17,26 @@ const (
...
+	DefaultMaxMarkerLogLines int = 500
 )
 
 // FTWConfiguration FTW global Configuration
 type FTWConfiguration struct {
-	LogFile             string          `koanf:"logfile"`
-	TestOverride        FTWTestOverride `koanf:"testoverride"`
-	LogMarkerHeaderName string          `koanf:"logmarkerheadername"`
-	RunMode             RunMode         `koanf:"mode"`
+	// Logfile is the path to the file that contains the WAF logs to check
```
theseion, line 28: Can the path be relative? If so, relative to the working directory?

config/types.go
```diff
@@ -17,14 +17,26 @@ const (
...
+	LogFile string `koanf:"logfile"`
+	// TestOverride holds the test overrides that will apply globally
+	TestOverride FTWTestOverride `koanf:"testoverride"`
+	// LogMarkerHeaderName is the name of the header that will be searched for on the logfile
```
theseion, line 32:
```suggestion
	// LogMarkerHeaderName is the name of the header that will be used by the test framework to mark positions in the log file
```

config/types.go
```diff
@@ -17,14 +17,26 @@ const (
...
+	LogMarkerHeaderName string `koanf:"logmarkerheadername"`
+	// RunMode stores the mode used to deciding the results of the test. See https://github.com/coreruleset/go-ftw#%EF%B8%8F-cloud-mode.
```
theseion, line 34:
```suggestion
	// RunMode stores the mode used to interpret test results. See https://github.com/coreruleset/go-ftw#%EF%B8%8F-cloud-mode.
```

theseion (changes requested):

config/types.go
fzipi: Yes. Even in the examples listed in the README you have a relative path example.

theseion (approved):

PR #125, opened 2022-12-11
Properly load config file

theseion:
The global variable holding the configuration was shadowed in the method that loades the configuration file. Hence, the configuration file was never applied.


fzipi (approved):
Nice catch! I wonder if we can test this somehow 🤔

theseion:
Yes, we can. Unfortunately, testing cobra is a bit involved and i wanted to get this fix in immediately. We can copy test setup for cobra from crs-toolchain.

PR #126, opened 2022-12-20
fix(goreleaser): invalid Slack blocks configuration

theseion:


fzipi:
Did we tested this? Maybe the release notes need the same `>` syntax?

theseion:
I couldn't test it because I don't have a token. But `goreleaser check` said the config was valid.
`>` is for multiline strings, where single newlines are replaced withs spaces and double newlines with a single newline. What do you mean by "Maybe the release notes need the same > syntax?".

fzipi (approved):
LGTM!

Issue #129, opened 2023-02-06
Conflict between missing/empty host CRS tests and setting "Host" header from dest_addr override


M4tteoP:
Hi!
I'm debugging CRS tests `920280-1` and `920290-1` that are failing against Coraza. It seems that there is a conflict between these specific tests (They are trying to send a request without a Host header or with an empty one) and the feature added in [feat: set "Host" header from dest_addr override](https://github.com/coreruleset/go-ftw/pull/63).
When a destination address override is provided, the feature replaces [the empty host field with the destination address](https://github.com/coreruleset/go-ftw/pull/63/files#diff-074224ff8d0619590a11bea64a759fe6be91e9b1f5d566011c1dd55a38454e49R368-R370) making the request a valid one.

It is not happening against Apache-Modsec (and CRS CI) because I think that destination overriding is not used there. A small ftw config like the following reproduces the issue:
```
logfile: '/Users/Repo/coreruleset/tests/logs/modsec2-apache/error.log'
testoverride:
  input:
    dest_addr: localhost
```

I may be missing something about #63, but I'm wondering why the `host` header is replaced only when `== ""`? I was expecting that this feature was mainly intended for replacing all the `Host: "localhost"` spread across the rules

I gave it a little go (just replacing `==` with `!=`, see [this branch](https://github.com/coreruleset/go-ftw/compare/main...M4tteoP:go-ftw:fix_host_override)): `920280-1` and `920290-1` would be fixed, but `["920270-4" "920350-1" "920350-3" "920350-4" "920350-5" "920350-6" "931130-9" "931130-11" "931130-15" "980170-2"]` start to fail.
It happens because, for example in test `920270-4`, a manipulated host is replaced by a legit one (`localhost%00` -> `localhost`).

Thanks for any input about it!

jcchavezs:
Ping @theseion

theseion:
You're right, running those tests with a `dest_addr` override will make them fail, since those rules check for missing and empty `Host` header. It is also true that we never run the test suite with any overrides.

I think what's missing is that the headers should also be applied from the overrides. But that's probably a different issue.

[...]

PR #132, opened 2023-02-21
feat(config): adds override_empty_host_header flag

M4tteoP:
The feature added in https://github.com/coreruleset/go-ftw/pull/63 replaces the empty host field with the destination address. It has conflicts with some CRS test that are trying to send on purpose empty or missing host headers.
It happens only when `dest_addr` override is in place, for this reason CRS tests against apache are still running smoothly.
This PR implements @theseion [idea](https://github.com/coreruleset/go-ftw/issues/129#issuecomment-1421347248), adding the `override_empty_host_header` flag that enables this feature, permitting, by default, to keep running CRS tests as intended. 
More context in the issue: https://github.com/coreruleset/go-ftw/issues/129.

Thanks the feedbacks!

tentatively closes https://github.com/coreruleset/go-ftw/issues/129

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,8 +296,20 @@ Other interesting functions you can use are: `randBytes`, `htpasswd`, `encryptAE
 Sometimes you have tests that work well for some platform combinations, e.g. Apache + modsecurity2, but fail for others, e.g. NGiNX + modsecurity3. Taking that into account, you can override test results using the `testoverride` config param. The test will be skipped, and the result forced as configured.
 
 Tests can be altered using four lists:
-- `input` allows you to override global parameters in tests. An example usage is if you want to change the `dest_addr` of all tests to point to an external IP or host
-  - it includes also the `override_empty_host_header` flag: enabling it, empty Host headers will be replaced with `dest_addr`
+- `input` allows you to override global parameters in tests. The following ones can be overridden:
+  - `dest_addr`: allows to point to an external IP or host
+  - `override_empty_host_header`: enabling this flag, empty Host headers will be replaced with `dest_addr`
+  - `port`: overrides the port number
+  - `protocol`: overrides the protocol
+  - `uri`: overrides the uri
+  - `version`: overrides the HTTP version. E.g. "HTTP/1.1"
+  - `headers`: overrides headers, the format is a map of strings
+  - `method`: overrides the method used to perform the request
+  - `data`: overrides data sent in the request
+  - `savecookie`: 
+  - `stopmagic`:
+  - `encodedrequest`: overrides base64 encoded request
+  - `rawrequest`: permits to provide a raw request. `method`, `uri` and `version` values will be ignored
 - `ignore` is for tests you want to ignore. You should add a comment on why you ignore the test
 - `forcepass` is for tests you want to pass unconditionally. You should add a comment on why you force to pass the test
 - `forcefail` is for tests you want to fail unconditionally. You should add a comment on why you force to fail the test
--- a/README.md
+++ b/README.md
@@ -297,8 +297,8 @@ Sometimes you have tests that work well for some platform combinations, e.g. Apa
 
 Tests can be altered using four lists:
 - `input` allows you to override global parameters in tests. The following ones can be overridden:
-  - `dest_addr`: allows to point to an external IP or host
-  - `override_empty_host_header`: enabling this flag, empty Host headers will be replaced with `dest_addr`
+  - `dest_addr`: overrides the destination address (accepts IP or hostname)
+  - `override_empty_host_header`: if true and `dest_addr` override is _also_ set, empty `Host` headers will be replaced with `dest_addr`
   - `port`: overrides the port number
   - `protocol`: overrides the protocol
   - `uri`: overrides the uri
@@ -306,8 +306,7 @@ Tests can be altered using four lists:
   - `headers`: overrides headers, the format is a map of strings
   - `method`: overrides the method used to perform the request
   - `data`: overrides data sent in the request
-  - `savecookie`: 
-  - `stopmagic`:
+  - `stopmagic`: prevent header autocompletion (currently sets `Connection: close` and `Content-Length` for requests with body data)
   - `encodedrequest`: overrides base64 encoded request
   - `rawrequest`: permits to provide a raw request. `method`, `uri` and `version` values will be ignored
 - `ignore` is for tests you want to ignore. You should add a comment on why you ignore the test
--- a/README.md
+++ b/README.md
@@ -297,6 +297,7 @@ Sometimes you have tests that work well for some platform combinations, e.g. Apa
 
 Tests can be altered using four lists:
 - `input` allows you to override global parameters in tests. An example usage is if you want to change the `dest_addr` of all tests to point to an external IP or host
+  - it includes also the `override_empty_host_header` flag: enabling it, empty Host headers will be replaced with `dest_addr`
 - `ignore` is for tests you want to ignore. You should add a comment on why you ignore the test
 - `forcepass` is for tests you want to pass unconditionally. You should add a comment on why you force to pass the test
 - `forcefail` is for tests you want to fail unconditionally. You should add a comment on why you force to fail the test
--- a/runner/run.go
+++ b/runner/run.go
@@ -379,7 +379,7 @@ func getRequestFromTest(testRequest test.Input) *ftwhttp.Request {
 
 // applyInputOverride will check if config had global overrides and write that into the test.
 func applyInputOverride(o config.FTWTestOverride, testRequest *test.Input) error {
-	overrides := o.Input
+	overrides := o.Overrides
 	if overrides.Port != nil {
 		testRequest.Port = overrides.Port
 	}
--- a/runner/run.go
+++ b/runner/run.go
@@ -383,14 +383,16 @@ func applyInputOverride(o config.FTWTestOverride, testRequest *test.Input) error
 	if overrides.Port != nil {
 		testRequest.Port = overrides.Port
 	}
-	if overrides.Headers.Get("Host") != "" {
+
+	if overrides.Headers != nil {
 		if testRequest.Headers == nil {
 			testRequest.Headers = ftwhttp.Header{}
 		}
-		if testRequest.Headers.Get("Host") == "" {
-			testRequest.Headers.Set("Host", overrides.Headers.Get("Host"))
+		for k, v := range overrides.Headers {
+			testRequest.Headers.Set(k, v)
 		}
 	}
+
 	if overrides.DestAddr != nil {
 		testRequest.DestAddr = overrides.DestAddr
 		if testRequest.Headers == nil {
--- a/runner/run.go
+++ b/runner/run.go
@@ -388,7 +388,7 @@ func applyInputOverride(o config.FTWTestOverride, testRequest *test.Input) error
 		if testRequest.Headers == nil {
 			testRequest.Headers = ftwhttp.Header{}
 		}
-		if testRequest.Headers.Get("Host") == "" {
+		if overrides.OverrideEmptyHostHeader && testRequest.Headers.Get("Host") == "" {
 			testRequest.Headers.Set("Host", *overrides.DestAddr)
 		}
 	}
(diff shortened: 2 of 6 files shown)


jcchavezs (approved):

README.md
```diff
@@ -297,6 +297,7 @@ Sometimes you have tests that work well for some platform combinations, e.g. Apa
 
 Tests can be altered using four lists:
 - `input` allows you to override global parameters in tests. An example usage is if you want to change the `dest_addr` of all tests to point to an external IP or host
+  - it includes also the `override_empty_host_header` flag: enabling it, empty Host headers will be replaced with `dest_addr`
```
jcchavezs, line 300: It feels like a breaking change. Are we OK with that? Shall we do it the other way around? Just loud thoughts.

M4tteoP: Being an override, part of the optional `testoverride` options, I felt that the correct default option would be to not override. Writing a test with an empty Host, I'm expecting that the Host is indeed sent empty.
If it sounds bad, I can turn it into `not_override_empty_host_header`

theseion, line 300: I think we should explicitly document all available overrides. That will make this section clearer.

theseion: Yes, it would be breaking, but I agree with @M4tteoP. I only know of one person who uses this feature and I don't think it would be too much to ask to make a small adjustment. We'll increase the major version number, that should suffice as a signal. @fzipi what do you think?

runner/run_test.go
```diff
@@ -675,7 +675,7 @@ func TestFailedTestsRun(t *testing.T) {
 	assert.Equal(t, 1, res.Stats.TotalFailed())
 }
 
-func TestApplyInputOverrideSetHostFromDestAddr(t *testing.T) {
+func TestApplyInputOverrideEmptyHostHeaderSetHostFromDestAddr(t *testing.T) {
```
theseion, line 678: Add another test for the default case, where `OverrideEmptyHostHeader` has not been set.

test/types.go
```diff
@@ -5,18 +5,19 @@ import "github.com/coreruleset/go-ftw/ftwhttp"
...
+	URI                     *string        `yaml:"uri,omitempty" koanf:"uri,omitempty"`
+	Version                 *string        `yaml:"version,omitempty" koanf:"version,omitempty"`
+	Headers                 ftwhttp.Header `yaml:"headers,omitempty" koanf:"headers,omitempty"`
+	Method                  *string        `yaml:"method,omitempty" koanf:"method,omitempty"`
+	Data                    *string        `yaml:"data,omitempty" koanf:"data,omitempty"`
+	SaveCookie              bool           `yaml:"save_cookie,omitempty" koanf:"save_cookie,omitempty"`
+	StopMagic               bool           `yaml:"stop_magic" koanf:"stop_magic,omitempty"`
+	EncodedRequest          string         `yaml:"encoded_request,omitempty" koanf:"encoded_request,omitempty"`
+	RAWRequest              string         `yaml:"raw_request,omitempty" koanf:"raw_request,omitempty"`
+	OverrideEmptyHostHeader bool           `yaml:"override_empty_host_header,omitempty" koanf:"override_empty_host_header,omitempty"`
```
theseion, line 20: I wonder whether we should start using a separate struct for the overrides. The new option has no place in the `input` map of tests.

theseion (changes requested):

README.md
M4tteoP: Done, I tried to write a small description of each, but not sure about `savecookie` and `stopmagic` 🤔

README.md
```diff
@@ -296,7 +296,20 @@ Other interesting functions you can use are: `randBytes`, `htpasswd`, `encryptAE
 Sometimes you have tests that work well for some platform combinations, e.g. Apache + modsecurity2, but fail for others, e.g. NGiNX + modsecurity3. Taking that into account, you can override test results using the `testoverride` config param. The test will be skipped, and the result forced as configured.
 
 Tests can be altered using four lists:
-- `input` allows you to override global parameters in tests. An example usage is if you want to change the `dest_addr` of all tests to point to an external IP or host
+- `input` allows you to override global parameters in tests. The following ones can be overridden:
+  - `dest_addr`: allows to point to an external IP or host
+  - `override_empty_host_header`: enabling this flag, empty Host headers will be replaced with `dest_addr`
```
theseion, line 301:
```suggestion
  - `override_empty_host_header`: if true and `dest_addr` override is _also_ set, empty `Host` headers will be replaced with `dest_addr`
```

README.md
```diff
@@ -296,7 +296,20 @@ Other interesting functions you can use are: `randBytes`, `htpasswd`, `encryptAE
 Sometimes you have tests that work well for some platform combinations, e.g. Apache + modsecurity2, but fail for others, e.g. NGiNX + modsecurity3. Taking that into account, you can override test results using the `testoverride` config param. The test will be skipped, and the result forced as configured.
 
 Tests can be altered using four lists:
-- `input` allows you to override global parameters in tests. An example usage is if you want to change the `dest_addr` of all tests to point to an external IP or host
+- `input` allows you to override global parameters in tests. The following ones can be overridden:
+  - `dest_addr`: allows to point to an external IP or host
```
theseion, line 300:
```suggestion
  - `dest_addr`: overrides the destination address (accepts IP or hostname)
```

README.md
```diff
@@ -296,7 +296,20 @@ Other interesting functions you can use are: `randBytes`, `htpasswd`, `encryptAE
...
+  - `override_empty_host_header`: enabling this flag, empty Host headers will be replaced with `dest_addr`
+  - `port`: overrides the port number
+  - `protocol`: overrides the protocol
+  - `uri`: overrides the uri
+  - `version`: overrides the HTTP version. E.g. "HTTP/1.1"
+  - `headers`: overrides headers, the format is a map of strings
+  - `method`: overrides the method used to perform the request
+  - `data`: overrides data sent in the request
+  - `savecookie`: 
+  - `stopmagic`:
```
theseion, line 310: If we keep this option, we should rename it IMO (keep it for backward compatibility but add a sibling that has a better name and document that here). @fzipi?

```suggestion
  - `stopmagic`: prevent header autocompletion (currently sets `Connection: close` and `Content-Length` for requests with body data)
```

README.md
```diff
@@ -296,7 +296,20 @@ Other interesting functions you can use are: `randBytes`, `htpasswd`, `encryptAE
...
+  - `dest_addr`: allows to point to an external IP or host
+  - `override_empty_host_header`: enabling this flag, empty Host headers will be replaced with `dest_addr`
+  - `port`: overrides the port number
+  - `protocol`: overrides the protocol
+  - `uri`: overrides the uri
+  - `version`: overrides the HTTP version. E.g. "HTTP/1.1"
+  - `headers`: overrides headers, the format is a map of strings
+  - `method`: overrides the method used to perform the request
+  - `data`: overrides data sent in the request
+  - `savecookie`:
```
theseion, line 309: Not yet implemented. @fzipi any plans to do that?

theseion (changes requested):

README.md
theseion: Thanks a lot!

README.md
M4tteoP: Many thanks for the suggested changes Max! Being not implemented I temptatively removed it from the list

fzipi: I think we need to uncomment https://github.com/coreruleset/go-ftw/blob/94ab311941820255097b13fe53c6f7276e19613a/ftwhttp/request.go#L165-L170 and provide some tests.

theseion: Ah yes. In that case @M4tteoP, I propose to drop `savecooke` for now, as you did. We can add it to the documentation once cookie handling has been implemented.

README.md
fzipi: This is part of the test yaml, right? Then we cannot change it without a new version of the yaml spec, sorry.

M4tteoP: So should I switch it to `not_override_empty_host_header `?
I think that #63  broke the default behavior, and enabling the custom behavior of overriding empty headers via `override_empty_host_header` would restore it.
We would have `dest_addr` that just overrides the destination address and the `override_empty_host_header` flag that force to override also the empty host headers, without any implicit behavior.

theseion: I don't understand your answer @fzipi 😅. I vote for `override_empty_host_header` (i.e., the "breaking change").

README.md
fzipi: Yes `stopmagic` is clearly not a good name regarding its semantic. We can use a better one.

theseion: I have a draft PR waiting to rename stopmagic after this PR has been merged.

README.md
fzipi: I don't mind the breaking change, as this is only in go-ftw. We should bail with an error for users that had the previous behavior, right?

theseion: How would we do that? Wouldn't that still be a valid configuration (valid in the context of testing, not w.r.t. HTTP)?
- dest_addr override would be set
- Host header would be empty
- Host header override would be empty

If you wanted to test request spoofing, i.e., use a host that differs from the target address, then you can't raise an exception here, right?

test/types.go
fzipi: I also think we should separate the overrides, yes.

fzipi (commented):
If we move the overrides to a new struct it will simplify things in the future...

runner/run.go
```diff
@@ -379,16 +379,26 @@ func getRequestFromTest(testRequest test.Input) *ftwhttp.Request {
 
 // applyInputOverride will check if config had global overrides and write that into the test.
 func applyInputOverride(o config.FTWTestOverride, testRequest *test.Input) error {
-	overrides := o.Input
+	overrides := o.Overrides
 	if overrides.Port != nil {
 		testRequest.Port = overrides.Port
 	}
+
+	if overrides.Headers != nil {
```
M4tteoP, line 387: I added headers override support starting from @snake66's [PR](https://github.com/coreruleset/go-ftw/pull/144) (Thank you!). Specifically speaking about the Host header, the individual header override will take precedence over a host header specified in the test and over the OverrideEmptyHostHeader feature. I added a test about it.

test/types.go
```diff
@@ -19,6 +20,23 @@ type Input struct {
 	RAWRequest     string         `yaml:"raw_request,omitempty" koanf:"raw_request,omitempty"`
 }
 
+// Overrides represents the overridden inputs that have to be applied to tests
+type Overrides struct {
```
M4tteoP, line 24: Little attempt to move the overrides to a new struct. When it will be settled, I will open another PR for extending the override capabilities (all the ones that I have commented out are currently not enforced)

runner/run_test.go
```diff
@@ -742,11 +770,101 @@ func TestApplyInputOverrideSetHostFromDestAddr(t *testing.T) {
...
+
+	err := applyInputOverride(cfg.TestOverride, &testInput)
+	assert.NoError(t, err, "Failed to apply input overrides")
+
+	assert.Equal(t, overrideHost, *testInput.DestAddr, "`dest_addr` should have been overridden")
+
+	assert.NotNil(t, testInput.Headers, "Header map must exist after overriding `dest_addr`")
+
 	hostHeader := testInput.Headers.Get("Host")
 	assert.NotEqual(t, "", hostHeader, "Host header must be set after overriding `dest_addr`")
```
theseion, line 800:
```suggestion
	assert.NotEqual(t, "", hostHeader, "Host header must be set after overriding `dest_addr` and setting `override_empty_host_header` to `true`")
```

runner/run_test.go
```diff
@@ -742,11 +770,101 @@ func TestApplyInputOverrideSetHostFromDestAddr(t *testing.T) {
...
 	assert.Equal(t, overrideHost, hostHeader, "Host header must be identical to `dest_addr` after overrding `dest_addr`")
```
theseion, line 801:
```suggestion
	assert.Equal(t, overrideHost, hostHeader, "Host header must be identical to `dest_addr` after overriding `dest_addr` and setting `override_emtpy_host_header` to `true`")
```

theseion (changes requested):

theseion (approved):

theseion:
Thanks @M4tteoP, looks good. I've requested a review from @fzipi, just to be sure we're all on the same page.

fzipi (approved):
Thanks for putting this together! LGTM.

camelmasa:
Great work! If possible, Please release a new version 👍

theseion:
https://github.com/coreruleset/go-ftw/releases/tag/v0.4.9

camelmasa:
Thank you so much!

PR #133, opened 2023-02-23
fix: change default user agent

fzipi:
Update default user agent.

### DIFF
--- a/ftwhttp/client_test.go
+++ b/ftwhttp/client_test.go
@@ -64,7 +64,7 @@ func TestGetTrackedTime(t *testing.T) {
 		Version: "HTTP/1.1",
 	}
 
-	h := Header{"Accept": "*/*", "User-Agent": "go-ftw test agent", "Host": "localhost"}
+	h := Header{"Accept": "*/*", "User-Agent": "ftw test agent", "Host": "localhost"}
 
 	data := []byte(`test=me&one=two&one=twice`)
 	req := NewRequest(rl, h, data, true)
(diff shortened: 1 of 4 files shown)


theseion:
This change isn't necessary. I've already made a change to the Docker image that prevents the test request header from ever being matched. I rather need a release of the Docker image so that the pipeline uses it to run the tests.

M4tteoP:
Hi @theseion, could you please point me to these changes?
I'm facing the same problem with Coraza test updating CRS to the latest commit:
```
[client "127.0.0.1"] Coraza: Warning. Remote Command Execution: Unix Shell Code Found in REQUEST_HEADERS [file "@owasp_crs/REQUEST-932-APPLICATION-ATTACK-RCE.conf"] [line "3762"] [id "932237"] [rev ""] [msg "Remote Command Execution: Unix Shell Code Found in REQUEST_HEADERS"] [data "Matched Data: go found within REQUEST_HEADERS:User-Agent: go-ftw test agent"] [severity "critical"] [ver "OWASP_CRS/4.0.0-rc1"] [maturity "0"] [accuracy "0"] [tag "application-multi"] [tag "language-shell"] [tag "platform-unix"] [tag "attack-rce"] [tag "paranoia-level/3"] [tag "OWASP_CRS"] [tag "capec/1000/152/248/88"] [tag "PCI/6.5.2"] [hostname ""] [uri "/status/200"] [unique_id "LBPtyWgXzsulWOfYjOx"]
```

theseion:
We’ve updated the user agent used by go-ftw for internal requests. You’ll need to update go-ftw for that to take effect.

--

theseion:
Yes, sorry. I was in between things and didn't really think this though. Instead of changing the user agent I opted to change special rule for marker logging that go-ftw uses, see https://github.com/coreruleset/modsecurity-crs-docker/pull/119/files.

jcchavezs:
My 2p: I really think we should change the user agent here and not remove
rules.

theseion:
That change was necessary anyway because this wasn't the first time where the marker request was matched by a rule. And it's really not an issue because this only concerns the marker requests, nothing else.

We can still change the user agent if you like.

Issue #134, opened 2023-02-26
About 411 Length Required response


camelmasa:
CRS 942101-* tests received "411 Length Required" data. go-ftw sends `Content-Length` header if request data is not 0 byte for now.
https://github.com/coreruleset/go-ftw/blob/main/ftwhttp/header.go#L78-L80

I think the `Content-Length` header needed if request header's method is "POST".

Like this.
https://github.com/coreruleset/go-ftw/commit/04ec4209aef484f060b4edbd3ee6312240b9d0a7#diff-2c39b9ef80e191a7385eb8d54e378f12a1ee08e30e6682274902b68deb46bbbdR112-R114

If possible, Could we create a PR for this?

---

.ftw.yaml
```
testoverride:
  input:
    protocol: "https"
    dest_addr: "oursite.com"
    port: 443
mode: "cloud"
```

Also, Following is our command log.

```
go-ftw  run -t -d coreruleset/tests/regression/tests/REQUEST-942-APPLICATION-ATTACK-SQLI/ -i 942101 --connect-timeout 10s --trace
```

```
running 942101-6: 11:01AM DBG ftw/http: sending data:
POST /unittests@coreruleset.org"%20sleep(10.to_i)%20" HTTP/1.0
Connection: close
Host: oursite.com



11:01AM TRC ftw/http: sending data
11:01AM TRC ftw/http: receiving data
11:01AM TRC ftw/http: received data - "HTTP/1.0 411 Length Required\r\nContent-Type: text/html; charset=UTF-8\r\nReferrer-Policy: no-referrer\r\nContent-Length: 286\r\nDate: Sun, 26 Feb 2023 02:01:21 GMT\r\n\r\n\n<html><head>\n<meta http-equiv=\"content-type\" content=\"text/html;charset=utf-8\">\n<title>411 Length Required</title>\n</head>\n<body text=#000000 bgcolor=#ffffff>\n<h1>Error: Length Required</h1>\n<h2>POST requests require a <code>Content-length</code> header.</h2>\n<h2></h2>\n</body></html>\n"
💥 failed in 146.461167ms (RTT 158.922959ms)
```

theseion:
I agree, the `Content-Length` headers should be present even when the payload is empty. What server is responding to that request? Both httpd and nginx are fine without the header.

camelmasa:
Thanks for a comment.

> What server is responding to that request?

[...]

PR #135, opened 2023-02-26
fix: send Content-Length header if there is payload

camelmasa:
Fixed: https://github.com/coreruleset/go-ftw/issues/134

CRS 942101-* tests received "411 Length Required" data. go-ftw sends Content-Length header if request data is not 0 byte for now.
https://github.com/coreruleset/go-ftw/blob/main/ftwhttp/header.go#L78-L80

I think the Content-Length header needed if request header's method is "POST".

### DIFF
--- a/ftwhttp/request.go
+++ b/ftwhttp/request.go
@@ -106,8 +106,8 @@ func (r *Request) AddHeader(name string, value string) {
 // AddStandardHeaders does the following:
 //   - adds `Connection` header with `close` value (if not set) to improve performance
 //   - adds `Content-Length` header if payload size > 0 or the request method
-//      permits a body (the spec says that the client SHOULD send `Content-Length`
-//      in that case)
+//     permits a body (the spec says that the client SHOULD send `Content-Length`
+//     in that case)
 func (r *Request) AddStandardHeaders() {
 	r.headers.Add("Connection", "close")
 
--- a/ftwhttp/request.go
+++ b/ftwhttp/request.go
@@ -107,7 +107,6 @@ func (r *Request) AddHeader(name string, value string) {
 func (r *Request) AddStandardHeaders() {
 	r.headers.Add("Connection", "close")
 
-	// If there is data or POST method header, we add the length also
 	if len(r.data) > 0 || r.requestLine.Method == "POST" {
 		r.headers.Add("Content-Length", strconv.Itoa(len(r.data)))
 	}
--- a/ftwhttp/request.go
+++ b/ftwhttp/request.go
@@ -105,7 +105,6 @@ func (r *Request) AddHeader(name string, value string) {
 //
 // This will add Content-Length and the proper Content-Type
 func (r *Request) AddStandardHeaders() {
-	// For better performance, we always close the connection (unless otherwise)
 	r.headers.Add("Connection", "close")
 
 	// If there is data or POST method header, we add the length also
--- a/ftwhttp/request.go
+++ b/ftwhttp/request.go
@@ -14,6 +14,8 @@ import (
 	"github.com/coreruleset/go-ftw/utils"
 )
 
+var methodsWithBodyRegex = regexp.MustCompile(`^POST|PUT|PATCH|DELETE$`)
+
 // ToString converts the request line to string for sending it in the wire
 func (rl RequestLine) ToString() string {
 	return fmt.Sprintf("%s %s %s\r\n", rl.Method, rl.URI, rl.Version)
@@ -114,13 +116,9 @@ func (r *Request) AddStandardHeaders() {
 		r.headers.Add("Connection", "close")
 	}
 
-	if len(r.data) > 0 || methodsWithBodyRegex().MatchString(r.requestLine.Method) {
+	if len(r.data) > 0 || methodsWithBodyRegex.MatchString(r.requestLine.Method) {
 		r.headers.Add("Content-Length", strconv.Itoa(len(r.data)))
 	}
-}
-
-func methodsWithBodyRegex() *regexp.Regexp {
-	return regexp.MustCompile(`^POST|PUT|PATCH|DELETE$`)
 }
 
 // isRaw is a helper that returns true if raw or encoded data
--- a/ftwhttp/request.go
+++ b/ftwhttp/request.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


ftwhttp/request.go
```diff
@@ -103,8 +104,14 @@ func (r *Request) AddHeader(name string, value string) {
 // AddStandardHeaders adds standard headers to the request, if they don't exist
 //
 // This will add Content-Length and the proper Content-Type
-func (r *Request) AddStandardHeaders(size int) {
-	r.headers.AddStandard(size)
+func (r *Request) AddStandardHeaders() {
+	// For better performance, we always close the connection (unless otherwise)
```
theseion, line 108:
```suggestion
```

ftwhttp/request.go
```diff
@@ -103,8 +104,14 @@ func (r *Request) AddHeader(name string, value string) {
...
+	r.headers.Add("Connection", "close")
```
theseion, line 109: Set only if the `Connection` header isn't already present.

ftwhttp/request.go
```diff
@@ -103,8 +104,14 @@ func (r *Request) AddHeader(name string, value string) {
...
+
+	// If there is data or POST method header, we add the length also
+	if len(r.data) > 0 || r.requestLine.Method == "POST" {
```
theseion, line 112: Define `methodsWithBodyRegex` as a constant with `regexp.MustCompile("^POST|PUT|PATCH|DELETE$")`.

```suggestion
	if len(r.data) > 0  || methodsWithBodyRegex.matchString(r.requestLine) {
```

ftwhttp/request.go
```diff
@@ -103,8 +104,14 @@ func (r *Request) AddHeader(name string, value string) {
 // AddStandardHeaders adds standard headers to the request, if they don't exist
 //
 // This will add Content-Length and the proper Content-Type
-func (r *Request) AddStandardHeaders(size int) {
-	r.headers.AddStandard(size)
+func (r *Request) AddStandardHeaders() {
+	// For better performance, we always close the connection (unless otherwise)
+	r.headers.Add("Connection", "close")
+
+	// If there is data or POST method header, we add the length also
```
theseion, line 111:
```suggestion
```

ftwhttp/request.go
```diff
@@ -103,8 +104,14 @@ func (r *Request) AddHeader(name string, value string) {
 // AddStandardHeaders adds standard headers to the request, if they don't exist
 //
 // This will add Content-Length and the proper Content-Type
```
theseion, line 106:
```suggestion
// AddStandardHeaders does the following:
//   - adds `Connection` header with `close` value (if not set) to improve performance
//   - adds `Content-Length` header if payload size > 0 or the request method
//      permits a body (the spec says that the client SHOULD send `Content-Length`
//      in that case)
```

theseion (changes requested):
Thanks!

camelmasa:
@theseion Got it. Thanks for reviewing. I'll update them then.

ftwhttp/request.go
camelmasa: Updated by https://github.com/coreruleset/go-ftw/pull/135/commits/84028c84b876ff6c60490572a97c5a946d12868c

ftwhttp/request.go
camelmasa: Updated by https://github.com/coreruleset/go-ftw/pull/135/commits/a6e2e79d88d9b9d7d6904bd5c131a9b7aab54e6a

camelmasa:
@theseion Okay, I updated them. How is this?

ftwhttp/request.go
```diff
@@ -102,9 +104,23 @@ func (r *Request) AddHeader(name string, value string) {
...
+		r.headers.Add("Connection", "close")
+	}
+
+	if len(r.data) > 0 || methodsWithBodyRegex().MatchString(r.requestLine.Method) {
+		r.headers.Add("Content-Length", strconv.Itoa(len(r.data)))
+	}
+}
+
+func methodsWithBodyRegex() *regexp.Regexp {
+	return regexp.MustCompile(`^POST|PUT|PATCH|DELETE$`)
```
theseion, line 123: I'd rather have this defined as a constant at file scope. It's not a huge deal but this regex will be applied many times during a full test run. Thus I'd like to cache compilation.

```go
var methodsWithBodyRegex = regexp.MustCompile(`^POST|PUT|PATCH|DELETE$`)
```

theseion (changes requested):

theseion:
Thanks! Almost perfect.

ftwhttp/request.go
camelmasa: I see! Good catch. 👍

camelmasa: @theseion Updated https://github.com/coreruleset/go-ftw/pull/135/commits/4bfdaca220be250a5e00860177778608ace60aad

jcchavezs:
> That's great, thanks. Would you mind adding 6 tests for that change? One for each of the methods with body, one for GET, and one for GET + non-empty payload (this should set the Content-Length header regardless of the method)?

Does it make a difference actually? I would just go for GET+e.g.PUT with/without body and move the regex verb check into a meaningful private method which could be tests if we really want to get 100% coverage.

Side note: What about `OPTIONS` verb?

theseion:
> Does it make a difference actually (to add 6 tests)? I would just go for GET+e.g.PUT with/without body and move the regex verb check into a meaningful private method which could be tests if we really want to get 100% coverage. I am coming from the fact that we should test in isolation as much as possible and testing all 4 verbs does not seem to add more value.

I'm not a Go expert, so I'm not very familiar with idioms and best practices. My reasoning for testing each verb is that I want to know when one of them disappears. There may also come a point where we have to extend the list (e.g. for WebDAV verbs) and then I'd want an additional test for each verb.
I'm not looking for 100% coverage, that doesn't add much value, IMO :).

> Side note: What about OPTIONS verb?

`OPTIONS` has no body, so the `Content-Length` header is unnecessary.

camelmasa:
@theseion I added tests https://github.com/coreruleset/go-ftw/pull/135/commits/e7b9771cb4ff9cafcdf8d9692fa5c60cf9cae580.

Also, I'm not Go expert too 😓

theseion (changes requested):
Looks good.
Can you add one last test? I want to make sure that it is possible to send data in a `GET` request. So the request method should be `GET`, the payload should not be empty and that should force the `Content-Length` header to be set.

camelmasa:
@theseion Okay, added tests https://github.com/coreruleset/go-ftw/pull/135/commits/f496432ca7878c3250b9525ddf2ed9eb96a06b8c

theseion (approved):

theseion:
Thanks for all the work @camelmasa!

camelmasa:
Thanks for reviewing!

PR #136, opened 2023-02-26
feat!: replace StopMagic with autocompleteHeaders

theseion:
StopMagic is still supported for backwards compatibility.

### DIFF
--- a/test/types.go
+++ b/test/types.go
@@ -17,11 +17,11 @@ type Input struct {
 	Method     *string        `yaml:"method,omitempty"`
 	Data       *string        `yaml:"data,omitempty"`
 	SaveCookie *bool          `yaml:"save_cookie,omitempty"`
-	// Deprecated: replaced with NoAutocompleteHeaders
-	StopMagic             *bool  `yaml:"stop_magic"`
-	NoAutocompleteHeaders *bool  `yaml:"no_autocomplete_headers"`
-	EncodedRequest        string `yaml:"encoded_request,omitempty"`
-	RAWRequest            string `yaml:"raw_request,omitempty"`
+	// Deprecated: replaced with AutocompleteHeaders
+	StopMagic           *bool  `yaml:"stop_magic"`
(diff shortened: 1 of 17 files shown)


fzipi:
Ugh, maybe `AutoCompleteHeaders` being true as the default (like now), and then setting it to false?

theseion:
That was my first impulse. However, there are no other fields that use values that differ from the type default yet, so I would have to switch to using constructors instead of `Input{}`. Might be a good idea, but I wanted to keep the PR small.

fzipi (commented):
Change looks good. But, we cannot merge this removing stop_magic now.

Once we have version 1.1 or 2.0 of the test spec, maybe.

theseion:
Ok. I'll rework it to retain `stop_magic` as legacy option.

fzipi:
@theseion How is that rework going on? Need help on this?

theseion:
If you want to, sure. I'm not getting much done at the moment.

theseion:
I've rewritten the PR to use `*bool` for all values in `Input` and `Overrides`. That makes it possible to detect missing values. Because of that I've also been able to change the name (and inversen the logic) to `AutocompleteHeaders`. `StopMagic` still works as before but `AutocompleteHeaders` has precedence.

test/types.go
```diff
@@ -66,9 +72,115 @@ type FTWTest struct {
...
 	} `yaml:"meta"`
 	Tests []Test `yaml:"tests"`
 }
+
+// ApplyInputOverride will check if config had global overrides and write that into the test.
+func ApplyInputOverrides(overrides *Overrides, input *Input) {
+	applySimpleOverrides(overrides, input)
+	applyDestAddrOverride(overrides, input)
+	applyHeadersOverride(overrides, input)
+	postProcessAutocompleteHeaders(overrides.AutocompleteHeaders, overrides.StopMagic, input)
```
fzipi, line 87: Way easier to read, thanks!

fzipi (approved):

Issue #137, opened 2023-03-10
respones_contains not checked if status specified


snake66:
We just have just moved our test suite from the original python ftw to this version, but one of the issues we have is that the `response_contains` key in the output section of a stage is not checked if the response status matches. This worked in the python based ftw.

**Example:**
```
    test_title: example
    stages:
      -
        stage:
          input:
            method: POST
            uri: /wp-admin/admin-ajax.php?action=<script>alert(1)</script>
            headers:
              Host: example.com
            data: '...'
          output:
            status: [403]
            response_contains: "rule 1234"
```

**Observed behaviour:**

The test passes, even if it's blocked by a different rule than the one specified.

**Expected behaviour:**

The test should fail.

theseion:
Yes, that sounds like a bug. Thanks for reporting it!

fzipi:
This should be fixed by the merged PR. Let's target a new release with this fix.

snake66:
Thanks for the quick fix! I have confirmed that this fixes the issue with our tests.

PR #139, opened 2023-03-12
fix: checkResult now matches all outputs required in the rule

M4tteoP:
Hi, I took a quick look at https://github.com/coreruleset/go-ftw/issues/137. As far as I understand, currently `checkResult` as soon as an assertion is true, returns `Success`, considering the rule as passed. This PR proposes to invert the logic, making sure that all the required assertions are not failing.
Not sure about these added `XRequired()` methods, maybe that could be merged into the `XAssert()` methods, but then it does not really become just an "assertion".

Feel free to close it if you have any better implementation in mind!

### DIFF
--- a/check/logs.go
+++ b/check/logs.go
@@ -8,10 +8,26 @@ func (c *FTWCheck) AssertNoLogContains() bool {
 	return false
 }
 
+// NoLogContainsRequired checks that the test requires no_log_contains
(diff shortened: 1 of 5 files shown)


fzipi (approved):
Thanks four your quick fix, @M4tteoP !

I think that the only thing we might want to add is a comment in the readme about this?

PR #144, opened 2023-03-27
Allow explicit override of Host header.

snake66:
Allows a default Host header to be set in the `testoverride` section of the config:

```
---
testoverride:
  dest_addr: some.example.org
  input:
    headers:
      Host: another.example.org
```

It will take precedense over `dest_addr` if it is specified as well. It will only be set if the input stage does not already have a host header set.

Falling back to using the DestAddr override if it's not specified.

### DIFF
--- a/runner/run.go
+++ b/runner/run.go
@@ -383,6 +383,14 @@ func applyInputOverride(o config.FTWTestOverride, testRequest *test.Input) error
 	if overrides.Port != nil {
 		testRequest.Port = overrides.Port
 	}
+	if overrides.Headers.Get("Host") != "" {
+		if testRequest.Headers == nil {
+			testRequest.Headers = ftwhttp.Header{}
+		}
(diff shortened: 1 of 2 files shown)


theseion:
Thanks @snake66. We already have a similar PR #132 that deals with Host header and `dest_addr`. What's missing int #132 to solve your issue is that `headers` are currently not applied as overrides, even though they should be. Would you mind basing your PR on #132 and changing it to apply header overrides? (ping @M4tteoP).

snake66:
Cool, no worries, I'll rebase on #132 and see if there's anything still missing there.

M4tteoP:
Hey! I'm on it, cherry-picked your commit, and working on applying broadly any header override. I will ping you soon on the other PR :)

snake66:
@M4tteoP I saw that. Just tested the branch with your additional changes and it works perfectly for me. :)

Closing this, then.

PR #153, opened 2023-05-24
fix: use full response to check in tests

### LINKED ISSUE #138: response_contains does not check the full response
The [ftw YAML format spec](https://github.com/coreruleset/ftw/blob/main/docs/YAMLFormat.md#response_contains) says:

> ### response_contains
>
> **Description:** Checks the entire response against the regular expression provided.

However, `go-ftw` only matches against the body of the response. This causes problems for tests that want to check that a a specific header is set by the firewall, or want to check responses that have no body.

fzipi:
Now tests will check against the full response (status, headers, body)

Fixes #138.


ftwhttp/response_test.go
```diff
@@ -126,3 +146,28 @@ func TestResponseWithCookies(t *testing.T) {
...
+	err = client.NewConnection(*d)
+
+	assert.NoError(t, err)
+
+	response, err := client.Do(*req)
+
+	assert.NoError(t, err)
+
+	assert.Contains(t, response.GetBodyAsString(), "User-Agent=[Go Tests]")
+	assert.NotContains(t, response.GetBodyAsString(), "X-Powered-By: [go-ftw]\n")
```
theseion, line 170: Shouldn't this be the same value as on the next line, where you check the full response?

```suggestion
	assert.NotContains(t, response.GetBodyAsString(), "X-Powered-By: go-ftw")
```

runner/run.go
theseion, line 1: Do you want to add an option to check response body only later? Otherwise, `GetBodyAsString` isn't of much use now.

theseion (changes requested):

ftwhttp/response_test.go
fzipi: Removed this and we are checking the full response.

theseion (approved):

PR #160, opened 2023-05-26
docs: update README

fzipi:
Add documentation for new "wait for" feature.


README.md
```diff
@@ -123,20 +123,31 @@ Usage:
...
+      --wait-delay duration                    Time to wait between retries for all wait operations. (default 1s)
+      --wait-for-connection-timeout duration   Http connection timeout, The timeout includes connection time, any redirects, and reading the response body. (default 3s)
+      --wait-for-expect-body-json string       Expect response body JSON pattern.
+      --wait-for-expect-body-regex string      Expect response body pattern.
+      --wait-for-expect-body-xpath string      Expect response body XPath pattern.
+      --wait-for-expect-header string          Expect response header pattern.
+      --wait-for-expect-status-code int        Expect response code e.g. 200, 204, ... .
+      --wait-for-host string                   Wait for host to be available before running tests.
+      --wait-for-insecure-skip-tls-verify      Skips tls certificate checks for the HTTPS request.
+      --wait-for-no-redirect                   Do not follow HTTP 3xx redirects.
```
theseion, line 149: I looked at this in the code quickly. The check function returns an error upon redirect. We might need to handle that properly. Could you add a test for that?

I also think the explanation needs to be more clear here, as it's not obvious what will happen. You can't really "wait for no redirect". I suspect that, in case of a redirect, that is the response that will be returned.

README.md
```diff
@@ -123,20 +123,31 @@ Usage:
...
+  -h, --help                                   help for run
+  -i, --include string                         include only tests matching this Go regexp (e.g. to include only tests beginning with "91", use "91.*").
+      --max-marker-log-lines int               maximum number of lines to search for a marker before aborting (default 500)
+      --max-marker-retries int                 maximum number of times the search for log markers will be repeated.
+                                               Each time an additional request is sent to the web server, eventually forcing the log to be flushed (default 20)
+  -o, --output string                          output type for ftw tests. "normal" is the default. (default "normal")
+      --read-timeout duration                  timeout for receiving responses during test execution (default 1s)
+      --show-failures-only                     shows only the results of failed tests
+  -t, --time                                   show time spent per test
+      --wait-delay duration                    Time to wait between retries for all wait operations. (default 1s)
```
theseion, line 140: What is the format for time specifications?

README.md
```diff
@@ -123,20 +123,31 @@ Usage:
...
+      --wait-for-connection-timeout duration   Http connection timeout, The timeout includes connection time, any redirects, and reading the response body. (default 3s)
+      --wait-for-expect-body-json string       Expect response body JSON pattern.
```
theseion, line 142: What is the pattern specification?

README.md
```diff
@@ -123,20 +123,31 @@ Usage:
...
+      --wait-for-expect-body-regex string      Expect response body pattern.
```
theseion, line 143: What is the pattern specification? Regex?

README.md
```diff
@@ -123,20 +123,31 @@ Usage:
...
+      --wait-for-expect-body-xpath string      Expect response body XPath pattern.
+      --wait-for-expect-header string          Expect response header pattern.
```
theseion, line 145: What is the pattern specification? Regex?

README.md
```diff
@@ -123,20 +123,31 @@ Usage:
...
+      --wait-for-expect-status-code int        Expect response code e.g. 200, 204, ... .
+      --wait-for-host string                   Wait for host to be available before running tests.
+      --wait-for-insecure-skip-tls-verify      Skips tls certificate checks for the HTTPS request.
```
theseion, line 148:
```suggestion
      --wait-for-insecure-skip-tls-verify      Skips TLS certificate checks for the HTTPS request.
```

theseion (changes requested):

README.md
fzipi: Yes.

README.md
fzipi: We are not using the `Check` func from the `Checker` interface. 🤷

README.md
fzipi: Added comment. We already had duration for other flags.

README.md
fzipi: Added comment and example.

README.md
fzipi: Added comment and example.

README.md
theseion: I disagree :) But I looked at the code again and the error is actually a special one from the Go HTTP library. It simply causes any redirect to be treated as the final response.

theseion (approved):

README.md
fzipi: Can you point to the code 🙏?

fzipi: This one? https://pkg.go.dev/net/http#Client.CheckRedirect

theseion: Yes, sorry.

PR #162, opened 2023-05-29
Reformat as testify

### LINKED ISSUE #106: Reformat test as testify/suite
In other tools we started using testify/suite that simplifies handling tests, setup and teardowns.

**Requirements**
- create suites for every test package
- move and simplify test creation

**Output**
- all tests changed to testify/suite
- all tests are passing

fzipi:
This reformat uses only testify for testing.

Major changes from the previous testing platform:
- Test names are used for getting files and or maps for tests that require it. This way we can use `BeforeTest` consistently for setting up tests.
- `text/template` is used for simplifying files handling. This probably could be optimized ever more.
- tests in the runner directory have been split
- added a new `httpsTestServer` for testing locally instead of calling remote hosts. this change also implied a change by adding `SetRootCAs` in the client.

Fixes #106.


config/config_test.go
```diff
@@ -5,171 +5,211 @@ import (
...
+func (s *fileTestSuite) AfterTest(_ string, _ string) {
+	if s.filename != "" {
+		err := os.Remove(s.filename)
+		s.NoError(err)
+		s.filename = ""
 	}
+}
+
+func (s *baseTestSuite) TestBaseUnmarshalText() {
+	var ftwRegex FTWRegexp
```
theseion, line 80:
```suggestion
	var ftwRegexp FTWRegexp
```

config/config_test.go
```diff
@@ -5,171 +5,211 @@ import (
...
+	err := ftwRegex.UnmarshalText([]byte("test"))
+	s.NoError(err)
+	s.NotNil(ftwRegex)
+	s.True(ftwRegex.MatchString("This is a test for unmarshalling"), "looks like we could not match string")
+}
+
+func (s *baseTestSuite) TestBaseNewFTWRegexpText() {
+	ftwRegex, err := NewFTWRegexp("test")
```
theseion, line 88:
```suggestion
	ftwRegexp, err := NewFTWRegexp("test")
```

config/config_test.go
```diff
@@ -5,171 +5,211 @@ import (
...
+
+	overrides := s.cfg.TestOverride.Overrides
+	s.NotNil(overrides.DestAddr, "Looks like we are not overriding destination address")
+	s.Equal("httpbingo.org", *overrides.DestAddr, "Looks like we are not overriding destination address")
+}
 
-	assert.NoError(t, err)
-	assert.NotNil(t, cfg)
+func (s *fileTestSuite) TestNewConfigBadConfig() {
+	// contents come from yamlBadConfig
```
theseion, line 129:
```suggestion
	// contents come from `bad` YAML config
```

config/config_test.go
```diff
@@ -5,171 +5,211 @@ import (
...
+	s.NoError(err)
+	s.NotNil(cfg)
 }
 
-func TestNewEnvConfigFromString(t *testing.T) {
-	cfg, err := NewConfigFromString(yamlConfig)
-	assert.NoError(t, err)
-	assert.NotNil(t, cfg)
+func (s *fileTestSuite) TestNewEnvConfigFromString() {
+	cfg, err := NewConfigFromString(testData["ok"])
```
theseion, line 150: Should test env, not string. Or should this test be deleted?

ftwhttp/client_test.go
```diff
@@ -1,63 +1,109 @@
...
+}
 
-	err = c.NewConnection(*d)
-	assert.NoError(t, err, "This should not error")
-	assert.Equal(t, "https", c.Transport.protocol, "Error connecting to example.com using https")
+func (s *clientTestSuite) SetupTest() {
+	var err error
+	s.client, err = NewClient(NewClientConfig())
+	s.NoError(err)
+	s.Nil(s.client.Transport, "Transport not expected to initialized yet")
```
theseion, line 27:
```suggestion
	s.Nil(s.client.Transport, "Transport not expected to be initialized yet")
```

config/config_test.go
```diff
@@ -5,171 +5,211 @@ import (
...
-		assert.Equal(t, "This test must be ignored", text, "Text doesn't match")
+func (s *fileTestSuite) BeforeTest(_, name string) {
+	var err error
+	s.filename, _ = utils.CreateTempFileWithContent(testData[name], "test-*.yaml")
+	s.cfg, err = NewConfigFromFile(s.filename)
+	s.NoError(err)
+	s.NotNil(s.cfg)
+}
+
+func (s *fileTestSuite) AfterTest(_ string, _ string) {
```
theseion, line 71: AFAICT, AfterTest and TearDownTest do the same thing, except that AfterTest has arguments. I suggest to use TearDownTest because we already use that and don't use the arguments.

ftwhttp/client_test.go
```diff
@@ -1,63 +1,109 @@
...
+			s.NoError(err)
+		}
+
+		_, err := w.Write(resp.Bytes())
+		s.NoError(err)
+	}))
+}
 
-	req := generateBaseRequestForTesting()
+func (s *clientTestSuite) httpsTestServer() {
```
theseion, line 50: Use one parameterized method instead of two with duplicated bodies?

ftwhttp/client_test.go
```diff
@@ -1,63 +1,109 @@
...
+	d, err := DestinationFromString(s.ts.URL)
+	s.NoError(err, "This should not error")
+	s.client.SetRootCAs(s.ts.Client().Transport.(*http.Transport).TLSClientConfig.RootCAs)
+	req := generateBaseRequestForTesting()
+	err = s.client.NewConnection(*d)
+	s.NoError(err, "This should not error")
+
+	response, err := s.client.Do(*req)
+
+	// I'm getting consistently 400 Bad Request from the server, so I'm commenting this out for now.
```
theseion, line 89: I don't understand what you commented out. Running `nc httpbin.org 80` works for me:

```
nc httpbin.org 80
UNEXISTENT /bad/path HTTP/1.4
Host: httpbin.org
User-Agent: curl/123
Accept: */*

HTTP/1.1 404 NOT FOUND
Date: Sun, 04 Jun 2023 19:39:36 GMT
Content-Type: text/html
Content-Length: 233
Connection: keep-alive
Server: gunicorn/19.9.0
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
```

runner/run_cloud_test.go
```diff
@@ -0,0 +1,115 @@
...
+	err = tmpl.Execute(testFileContents, vars)
+	s.NoError(err, "cannot execute template")
+	// get tests from file
+	s.ftwTests, err = test.GetTestsFromFiles(testFileContents.Name())
+	s.NoError(err, "cannot get tests from file")
+	// save the name of the temporary file so we can delete it later
+	s.tempFileName = testFileContents.Name()
+}
+
+func (s *runCloudTestSuite) AfterTest(_ string, _ string) {
```
theseion, line 77: AFAICT, AfterTest and TearDownTest do the same thing, except that AfterTest has arguments. I suggest to use TearDownTest because we already use that and don't use the arguments.

runner/run_input_override_test.go
```diff
@@ -0,0 +1,319 @@
...
+		if overrideConfigMap[name] == nil {
+			return "", errors.New("cannot get override config value: be sure the caller is a test function, and the key is correct")
+		}
+		if strings.Contains(key, ".") {
+			keyParts := strings.Split(key, ".")
+			return overrideConfigMap[name].(map[string]interface{})[keyParts[0]].(map[string]string)[keyParts[1]], nil
+		}
+		return overrideConfigMap[name].(map[string]interface{})[key].(string), nil
+	}
+	return "", errors.New("cannot caller function name")
```
theseion, line 114:
```suggestion
	return "", errors.New("failed to determine calling function")
```

theseion (changes requested):
Phew! That took longer than expected ;)

Thanks!

config/config_test.go
fzipi: done.

config/config_test.go
fzipi: Done.

config/config_test.go
fzipi: Done.

config/config_test.go
fzipi: Removed.

ftwhttp/client_test.go
fzipi: No, the problem is that this used to be the call but now we are using local calls to the `httpsTestServer`. Calling that server will give you 400. Updating the comment to it is more clear.

ftwhttp/client_test.go
fzipi: Sure.

runner/run_cloud_test.go
fzipi: You are right, removing func and moving cleanup to TearDown.

runner/run_input_override_test.go
fzipi: Done.

ftwhttp/client_test.go
fzipi: Done.

ftwhttp/client_test.go
```diff
@@ -1,63 +1,112 @@
...
+	// so I'm commenting this out for now.
+	// Example call used to test this:
+	// nc httpbin.org 80
+	// UNEXISTENT /bad/path HTTP/1.4
+	// Host: httpbin.org
+	// User-Agent: curl/7.88.1
+	// Accept: */*
+	//
+	// Will return HTTP/1.1 404 NOT FOUND from upstream.
+	// The same call to the httpsTestServer will return HTTP/1.1 400 Bad Request.
```
theseion, line 98: I debugged this. The server expects two things for this to work properly:
1. The HTTP version must conform to spec. Currently you transfer `1.4` but it must be `HTTP/1.4`
2. The `Host` header must be present

theseion (changes requested):

ftwhttp/client_test.go
fzipi: Genius. Thanks 💪

fzipi: Ok, still cannot make this fail. Somehow the outside query is making the request to fail, even if it is not failing. I mean, you just get a bad request response. I wonder how the previous assertion worked at all 🤔

theseion: Your test server writes `http.StatusOK`, so the last line of the test should be (notice that I also fixed the order of paramaeters):
```
s.Equal(http.StatusOK, response.Parsed.StatusCode, "Error in calling website")
```

Your test server just doesn't care about the path. You could change it to care if you really wanted a 404.

theseion (approved):

PR #175, opened 2023-08-07
fix: reduce log level when writing headers

theseion:
Multiple log entries are produced for each test, for each HTTP header written.


ftwhttp/header.go
```diff
@@ -102,7 +102,7 @@ func (h Header) WriteBytes(b *bytes.Buffer) (int, error) {
 	for _, key := range sorted {
 		// we want all headers "as-is"
 		s := key + ": " + h[key] + "\r\n"
-		log.Info().Msgf("Writing header: %s", s)
+		log.Debug().Msgf("Writing header: %s", s)
```
fzipi, line 105: Maybe using `Trace` instead?

fzipi (approved):
Lovely and quiet.

PR #176, opened 2023-08-07
fix: no longer negate `autocomplete_headers` option

theseion:
When the logic was inverted, as we renamed `stop_magic` to `autocomplete_headers`, we forgot to remove the logic inversion when passing the option to the request builder.


fzipi (changes requested):
I would say let's write a test to catch this?

theseion:
Yes... I was being lazy and you caught me 😅.

theseion:
Should be good now.

fzipi (approved):

PR #177, opened 2023-08-13
fix runner expected-error check potential panic

4lisalehi:
Not sure if I'm using and testing the runner package correctly or not, but since `expectedOutput.ExpectError` is a pointer to boolean, dereferencing it is inherently unsafe and panic-able and depending on the behavior and outputs of go-ftw tests yaml deserialization, logically, it could panic in some cases and I believe this small change makes it a bit safer.

If I'm missing any point here, I'd appreciate your clarifying it.

Thanks.


theseion (approved):

theseion:
Thanks @4lisalehi!

PR #208, opened 2023-11-09
feat: move test type to its own package

fzipi:
## what

- move internal test type to use external package

## why

- allow others to consume exported types
- easy versioning on the yaml schema


M4tteoP:
Currently depends on https://github.com/coreruleset/ftw-tests-schema/pull/5 branch. It is meant to point to schema v1.1 before merging

test/defaults.go
```diff
@@ -68,3 +82,11 @@ func (i *Input) GetRawRequest() ([]byte, error) {
 	}
 	return nil, nil
 }
+
+// GetAutocompleteHeaders returns the proper the autocompleteHeaders value, defaults to true
```
theseion, line 86:
```suggestion
// GetAutocompleteHeaders returns the autocompleteHeaders value, defaults to true
```

theseion (changes requested):

test/defaults.go
M4tteoP: Palinka's fault

theseion (approved):

PR #268, opened 2024-03-07
chore: add OpenSSF badge

fzipi:
I went all over the openssf badge process, so we might add the badge here.


theseion (commented):
Did you choose this badge on purpose? For the toolchain I used the score badge. Not sure which ones better. Or both?
I remember that the best practice badge requires an account, right? Which is one reason that I chose the score badge.

theseion:
Did you see my question above?

fzipi:
Done.

theseion (approved):

PR #274, opened 2024-03-14
chore: rename codeql workflow

theseion:
The CodeQL workflow is currently not being run and it is not clear why. Try renaming the workflow to force GH to run it.


theseion:
Looks like this might work, but we'll need to see what happens after the merge.

fzipi (approved):

PR #276, opened 2024-03-16
chore: run renovate only once per week to reduce noise

theseion:


dune73:
Does the renovate bot actions also include security relevant findings? If yes, I'd rather have the noise than the delay.

theseion:
Potentially, yes. But it really doesn't help to have the noise. Even if there's a relevant issue, we would still have to cut a release etc. IMO, there's no difference whether this occurs a week earlier or later, especially, since we're talking about a test tool here, not a public web server. Remember, that it's pretty much me and @fzipi carrying the weight and we also have other things to do.

dune73:
Agreed. It makes sense in such a situation.

fzipi (approved):

PR #281, opened 2024-03-21
feat: add client rate limits

### LINKED ISSUE #269: Throttle request speed option
Nearly all commercial WAFs have a "restrict number of requests per minute per IP" setting. Most should also have a separate "restrict number of requests per minute per session" setting.

It would be nice if we could throttle go-ftw to only send X requests per minute (especially for cloud mode). Otherwise go-ftw will hit the X/min limitation and will only produce false positives after that.

fzipi:
Fixes #269.


ftwhttp/client_test.go
```diff
@@ -211,3 +226,33 @@ func (s *clientTestSuite) TestNewOrReusedConnectionReusesTransport() {
 
 	s.Equal(begin, s.client.Transport.duration.begin, "Transport must not be reinitialized when reusing connection")
 }
+
+// TestClientRateLimits tests the rate limiter functionality of the client. Test should take at least 5 seconds to run.
```
theseion, line 230:
```suggestion
// TestClientRateLimits tests the rate limiter functionality of the client. Test should take at least 3 seconds to run.
```

ftwhttp/types.go
```diff
@@ -18,6 +20,8 @@ type ClientConfig struct {
 	ReadTimeout time.Duration
 	// RootCAs is the set of root CA certificates that is used to verify server
 	RootCAs *x509.CertPool
+	// RateLimiter is the rate limiter to use for requests.
+	Ratelimiter *rate.Limiter
```
theseion, line 24:
```suggestion
	RateLimiter *rate.Limiter
```

cmd/run.go
```diff
@@ -54,6 +54,7 @@ func NewRunCommand() *cobra.Command {
 	runCmd.Flags().Duration("wait-for-connection-timeout", http.DefaultConnectionTimeout, "Http connection timeout, The timeout includes connection time, any redirects, and reading the response body.")
 	runCmd.Flags().Bool("wait-for-insecure-skip-tls-verify", http.DefaultInsecureSkipTLSVerify, "Skips tls certificate checks for the HTTPS request.")
 	runCmd.Flags().Bool("wait-for-no-redirect", http.DefaultNoRedirect, "Do not follow HTTP 3xx redirects.")
+	runCmd.Flags().DurationP("rate-limit", "r", 0, "Rate limit the requests to the server. 0 is the default, and means unlimited.")
```
theseion, line 57:
```suggestion
	runCmd.Flags().DurationP("rate-limit", "r", 0, "Limit the request rate to the server to 1 request per specified duration. 0 is the default, and disables rate limiting.")
```

theseion (changes requested):
Not entirely sure about the use case, but would it help to also add an option for jitter? I seem to recall that some web servers don't like it if requests are too regular, but I may be mistaken.

fzipi:
For now the use case doesn't mention jitter. I would go with this now and see the reaction. Jitter also doesn't come with the package, so we must implement it...

PR #289, opened 2024-04-06
feat: run CRS regression tests

theseion:


fzipi (commented):
Why a generic tool should depend on CRS for waf testing?

theseion:
It shouldn't. But the test will alert us to breaking changes. Since CRS cannot merge when tests fail, we know that the main branch is supposed to work.

I've thought about implementing this test on the CRS side, but I figured it would help us more if we saw a test failure during a PR than after having merged.

fzipi:
Then my proposal is to switch to controlled deployment in the CRS gha instead. Let's add the tool version in the ci/cd on the CRS side. That way the control is on the other side, and we don't depend here on a repo that, while related, we do not depend on it. And also, things can be merged that break this repo. 🤷

PR #298, opened 2024-05-11
fix: properly handle multiple IDs in positive log match

theseion:
The implementation for positive match of a list of IDs only required a single ID to match instead of all.

This commit also introduces memoization to minimize IO when searching the log.

Also fixed pre-commit config.


fzipi (approved):
Well, I hope that my IDE doesn't change imports back again. This is going to bite us in the future probably...

PR #310, opened 2024-06-01
feat: improve file loading robustness

theseion:
Until now, go-ftw didn't know what to do with YAML files that were not test files. With the new platform override files (also YAML), it is very easy to also load them when globbing. This commit improves the output for such issues and continues execution, ignoring any files that can't be loaded.


fzipi (approved):
Nice feature!

PR #313, opened 2024-06-08
fix: check all matches when looking for rule IDs

theseion:


fzipi:
Should we add some tests here?

theseion:
Done.

waflog/read.go
```diff
@@ -28,16 +28,21 @@ func (ll *FTWLogLines) TriggeredRules() []uint {
...
+		match := regex.FindAllSubmatch(line, -1)
 		if match != nil {
 			log.Trace().Msgf("ftw/waflog: Found %s at %s", regex.String(), line)
-			submatch := string(match[1])
-			ruleId, err := strconv.ParseUint(submatch, 10, 0)
-			if err != nil {
-				log.Error().Msgf("Failed to parse uint from %s", submatch)
-				continue
+			for _, nextMatch := range match {
+				submatch := string(nextMatch[1])
```
huberts90, line 35: Could you please check if the `len(nextMatch) > 0` firstly?

waflog/read.go
```diff
@@ -28,16 +28,21 @@ func (ll *FTWLogLines) TriggeredRules() []uint {
...
+				if len(submatch) == 0 {
+					submatch = string(nextMatch[2])
```
huberts90, line 37: Could you please check if the `len(nextMatch) > 1` firstly?

waflog/read.go
theseion: Unnecessary. When the expression matches successfully, `nextMatch` cannot be empty. The submatches are not optional.

waflog/read.go
theseion: Same as above.

waflog/read.go
theseion: I've replaced the explicit accesses with a loop. Thanks for the hint.

PR #314, opened 2024-06-08
refactor: reduce memory thrashing in log search

theseion:


fzipi (dismissed):

waflog/read.go
```diff
@@ -16,16 +16,17 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+var ruleIdsSet = make(map[uint]bool, 15)
```
fzipi, line 19: Can we use a constant with a proper name, instead of this arbitrary 15?

fzipi (commented):
Nice, slices!

fzipi (approved):
Thanks!

PR #320, opened 2024-06-18
Fix: Verify executed tests taking include/exclude option into account

huberts90:


runner/run_test.go
```diff
@@ -312,6 +316,7 @@ func (s *runTestSuite) TestRunTests_Run() {
 			Output:  output.Quiet,
 		}, s.out)
 		s.Require().NoError(err)
+		s.Len(res.Stats.Success, 5, "failed to exclude test")
```
theseion, line 319: Based on the original IDs, this testshould probably exclude all tests (leading 0).

theseion (changes requested):

runner/run_test.go
huberts90: As far as I understand, `include` [precedes](https://github.com/coreruleset/go-ftw/blob/main/runner/run.go#L262) `exclude` and having a regex like `1*` matches `zero` - `unlimited`. 
What is the expected result now, though?

theseion: The expressions are useless. `1*` will match everything, because the match can be of length zero. So, firstly, you need to change the expressions to `-1.*`, etc., secondly, the test needs to assert 4 successful and 1 skipped test run.

huberts90: Exactly, the tests have been extended with verification of `skipped` lists.

theseion (approved):

theseion:
We have the policy that all commits have to be signed. I'm going to bypass that policy in this case. Please set up [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

PR #331, opened 2024-07-09
Header getter/setter should be case in-sensitive

huberts90:


ftwhttp/header.go
```diff
@@ -34,7 +35,9 @@ func (w stringWriter) WriteString(s string) (n int, err error) {
 
 // Add adds the key, value pair to the header.
 // It appends to any existing values associated with key.
+// The key is case in-sensitive
```
theseion, line 38:
```suggestion
// The key is case-insensitive
```

ftwhttp/header.go
```diff
@@ -43,35 +46,37 @@ func (h Header) Add(key, value string) {
 // Set sets the header entries associated with key to
 // the single element value. It replaces any existing
 // values associated with key.
+// The key is case in-sensitive
```
theseion, line 49:
```suggestion
// The key is case-insensitive
```

ftwhttp/header.go
```diff
@@ -43,35 +46,37 @@ func (h Header) Add(key, value string) {
...
 func (h Header) Set(key, value string) {
-	h[key] = value
+	h[canonicalHeaderKey(key)] = value
 }
 
 // Get gets the first value associated with the given key.
-// It is case insensitive;
 // If there are no values associated with the key, Get returns "".
+// The key is case in-sensitive
```
theseion, line 56:
```suggestion
// The key is case-insensitive
```

ftwhttp/header.go
```diff
@@ -43,35 +46,37 @@ func (h Header) Add(key, value string) {
...
 	}
-	v := h[key]
+	v := h[canonicalHeaderKey(key)]
 
 	return v
 }
 
 // Value returns the value associated with the given key.
-// It is case insensitive;
+// The key is case in-sensitive
```
theseion, line 67:
```suggestion
// The key is case-insensitive
```

ftwhttp/header.go
```diff
@@ -34,7 +35,9 @@ func (w stringWriter) WriteString(s string) (n int, err error) {
 
 // Add adds the key, value pair to the header.
 // It appends to any existing values associated with key.
+// The key is case in-sensitive
 func (h Header) Add(key, value string) {
```
theseion, line 39: This method doesn't do what its comment says. I'll open a separate issue for that.

ftwhttp/header.go
```diff
@@ -43,35 +46,37 @@ func (h Header) Add(key, value string) {
...
 	if h == nil {
 		return ""
 	}
 
-	return h[key]
+	return h[canonicalHeaderKey(key)]
 }
 
 // Del deletes the value associated with key.
+// The key is case in-sensitive
```
theseion, line 77:
```suggestion
// The key is case-insensitive
```

ftwhttp/header.go
```diff
@@ -138,3 +143,8 @@ func (h Header) getSortedHeadersByName() []string {
 
 	return keys
 }
+
+// canonicalHeaderKey transforms given to the canonical form
+func canonicalHeaderKey(key string) string {
```
theseion, line 148:
```suggestion
func canonicalKey(key string) string {
```

ftwhttp/header.go
```diff
@@ -138,3 +143,8 @@ func (h Header) getSortedHeadersByName() []string {
 
 	return keys
 }
+
+// canonicalHeaderKey transforms given to the canonical form
```
theseion, line 147:
```suggestion
// canonicalKey transforms given to the canonical form
```

ftwhttp/header_test.go
```diff
@@ -111,6 +111,11 @@ func (s *headerTestSuite) TestHeaderSetGet() {
 	h.Add("Other", "Value")
 	value := h.Get("Other")
 	s.Equalf("Value", value, "got: %s, want: %s\n", value, "Value")
+
+	// Case in-sensitive
```
theseion, line 115: Make this a separate test case.

ftwhttp/header_test.go
```diff
@@ -124,6 +129,13 @@ func (s *headerTestSuite) TestHeaderDel() {
 			s.Equalf("", value, "#%d: got: %s, want: %s\n", i, value, "")
 		}
 	}
+
```
theseion, line 132: Make this a separate test case.

ftwhttp/header_test.go
theseion, line 1: Please add one test per method.

theseion (changes requested):

ftwhttp/header.go
theseion: https://github.com/coreruleset/go-ftw/issues/332

ftwhttp/header_test.go
```diff
@@ -104,20 +104,54 @@ func (s *headerTestSuite) TestHeaderWriteString() {
...
 }
 
-func (s *headerTestSuite) TestHeaderSetGet() {
+func (s *headerTestSuite) TestHeaderAdd() {
 	h := Header{
 		"Custom": "Value",
 	}
 	h.Add("Other", "Value")
-	value := h.Get("Other")
+	value := h["Other"]
```
theseion, line 112: Use `Get` for access in all tests. You shouldn't test the internal representation.

ftwhttp/header.go
```diff
@@ -35,9 +35,9 @@ func (w stringWriter) WriteString(s string) (n int, err error) {
 
 // Add adds the key, value pair to the header.
 // It appends to any existing values associated with key.
-// The key is case in-sensitive
+// The key is case-insensitive
 func (h Header) Add(key, value string) {
-	key = canonicalHeaderKey(key)
+	key = canonicalKey(key)
```
theseion, line 40: Sorry, I just realised that using `canonicalKey` doesn't work. We need to retain the original casing of the header name. Instead, `Get` should use `canonicalKey` when comparing keys for retrieval.

theseion (changes requested):

ftwhttp/header_test.go
huberts90: We both agreed that a unit test should check one method to easily identify what failed if the test fails. Do we want to strictly adhere to this principle or not?  For me we do not test the internal representation, we verify if `Add`/`Get` worked as expected. I am eager to know your opinion about that.

ftwhttp/header.go
huberts90: I do not understand how it would work if only `Get` retrieves data by the "canonicalized" key. The original [header.go](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/http/header.go;l=30-58) works in a different way if we want to model it after them

ftwhttp/header_test.go
theseion: By using `h["X"]`, you are verifying the internal representation because you rely on the knowledge that the canonical form of `x` is `X`. In theory, we might want to change the internal representation, without changing the API, then these tests will break.

ftwhttp/header.go
theseion: There are at least two possible ways to accomplish this:
1. iterate over the keys of the map, canonicalise each key and compare it to the canonicalised version of the search key
2. modify the internal state, e.g., use a map `(canonical(key)) -> (map(header-name, header-value))`, then you can use an efficient lookup to find all the headers that reduce to the canonical form

theseion (approved):

theseion:
Thanks!

PR #340, opened 2024-08-03
fix: overridden tests displayed as failing

M4tteoP:
### Summary:
Overridden results were added to the stats twice: once by the overridden logic that was correctly setting them as overridden tests, and once by `RunTest` that was adding the overridden test using the result of the previous run.

### Full story:
Bumped into the following output where `920290-1` was displayed as both an ignored and failed. 
```
--- FAIL: TestFTW (14.27s)
    coreruleset_test.go:284: 7 ignored tests: [920100-5 920100-8 920270-4 920290-1 920290-4 920430-8 930110-7]
    coreruleset_test.go:288: 8 failed tests: [920274-1 920280-3 920290-1 920430-3 920430-5 920430-9 920610-2 920620-1]
```
When an overridden test was detected in `RunStage`, the result addition was not delegated to the `RunTest` function, but directly called there (`addResultToStats`). Once returned, `addResultToStats` was executed once again, storing the value of `runContext.Result` which, in case of an overridden test, contained the result of the previous run.

All the other overrides were anticipated by a successful test, therefore they did not show the bug.

### DIFF
--- a/runner/run.go
+++ b/runner/run.go
@@ -148,9 +148,9 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 		return err
 	}
 
-	// Do not even run test if result is overridden. Just use the override and display the overridden result.
+	// Do not even run test if result is overridden. Directly set and display the overridden result.
 	if overridden := overriddenTestResult(ftwCheck, &testCase); overridden != Failed {
-		runContext.Stats.addResultToStats(overridden, &testCase)
+		runContext.Result = overridden
 		displayResult(&testCase, runContext, overridden, time.Duration(0), time.Duration(0))
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


runner/run.go
```diff
@@ -148,9 +148,9 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 		return err
 	}
 
-	// Do not even run test if result is overridden. Just use the override and display the overridden result.
+	// Do not even run test if result is overridden. Directly set and display the overridden result.
 	if overridden := overriddenTestResult(ftwCheck, &testCase); overridden != Failed {
-		runContext.Stats.addResultToStats(overridden, &testCase)
+		runContext.Result = overridden
```
M4tteoP, line 153: This is the one line fix. The result is not added here to the stats, but is updated with the overridden result. The caller will then add it to the stats like any other executed test.

fzipi (approved):
Nice find!

M4tteoP:
Ops, rebased, and signed the commit ✍️

fzipi (approved):

theseion (approved):

PR #380, opened 2024-11-02
fix: renamed '-d' option for quantitative testing

dune73:


M4tteoP (approved):

M4tteoP:
I had to squash the commits to have the commit with a verified signature, so the CI is happy

PR #382, opened 2024-11-02
quantitative: runs with `DetectionOnly`, fixes `--payload`

M4tteoP:
- Adds `q` as `quantitative` alias
- Runs quantitative testing with `DetectionOnly`: detection is not interrupted anymore if phase 1 rules match.
- code Refactor to better fit providing the `--payload` parameter
- Fixes `--payload`


fzipi (approved):
🚀

PR #394, opened 2024-11-09
feat: rename `buildRequest` to `BuildRequest` to make it exported

TimDiam0nd:
I'm aiming to import the ftwhttp package into another Go project and call the `buildRequest` function directly. However, since `buildRequest` is unexported (lowercased), it’s inaccessible from outside the package.

Renaming `buildRequest` to `BuildRequest` would make it an exported function, allowing me to use it outside of go-ftw

### DIFF
--- a/ftwhttp/connection.go
+++ b/ftwhttp/connection.go
@@ -84,7 +84,7 @@ func (c *Connection) receive() (io.Reader, error) {
 // Request will use all the inputs and send a raw http request to the destination
 func (c *Connection) Request(request *Request) error {
 	// Build request first, then connect and send, so timers are accurate
-	data, err := buildRequest(request)
+	data, err := BuildRequest(request)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


theseion:
Interesting! `connection.Request` is exported, so you could use that instead. Could you explain your use case? Maybe we can export a function that makes more sense, since `buildRequest` has a lot of logic.

TimDiam0nd:
I'd like to do some further logic/control on how/where the request is sent (outside Go), thus i just want the request object to then further manipulate and use.

theseion:
Thanks. I'd prefer a better mechanism for exposing that functionality, e.g., a default implementation of an interface, but that would require quite a bit of work. I suppose this is good for now.

theseion (approved):

theseion:
@TimDiam0nd please consider setting up commit signing. We usually require signed commits.

PR #395, opened 2024-11-10
chore: upgrade go-getter to v2

theseion:


theseion:
Added the fix to `File` here because some of the Windows tests were failing due to a race condition when caching.

fzipi (approved):

PR #405, opened 2024-11-24
feat: don't trace log when skipping lines

theseion:
When using `-n` to run a single line only, trace logging will log every single line. Stop doing that.


theseion:
f

M4tteoP (approved):
LGTM, I thought there was no limit on being verbose with Trace 😅

PR #410, opened 2024-12-06
Extend ID Regex to support escaped quotes

M4tteoP:
Address https://github.com/coreruleset/coreruleset/issues/3239.
This PR proposes to extend the regex looking for IDs to support also IDs coming with escaped quotes.
JSON log lines such as the following one are not supported:
```
{"time":"2024-12-06T11:13:12.250069+01:00","level":"ERROR","msg":"[client \"127.0.0.1\"] Coraza: Warning. Request content type charset is not allowed by policy [file \"@owasp_crs/REQUEST-920-PROTOCOL-ENFORCEMENT.conf\"] [line \"2753\"] [id \"920480\"] [rev \"\"] [msg \"Request content type charset is not allowed by policy\"] [data \"|garbage|\"] [severity \"critical\"] [ver \"OWASP_CRS/4.7.0\"] [maturity \"0\"] [accuracy \"0\"] [tag \"application-multi\"] [tag \"language-multi\"] [tag \"platform-multi\"] [tag \"attack-protocol\"] [tag \"paranoia-level/1\"] [tag \"OWASP_CRS\"] [tag \"capec/1000/255/153\"] [tag \"PCI/12.1\"] [hostname \"127.0.0.1\"] [uri \"/\"] [unique_id \"UGAYIvYnOAtQVNqhxNg\"]","severity":"critical"}
```
As described in the original issue, the problem is `[id \"920480\"]`, but now as mentioned in https://github.com/coreruleset/coreruleset/issues/3239#issuecomment-2096775153, we can just extend the regex to make it flexible enough to support also this case.

### DIFF
--- a/waflog/read.go
+++ b/waflog/read.go
@@ -28,7 +28,7 @@ func (ll *FTWLogLines) TriggeredRules() []uint {
 	ll.triggeredRulesInitialized = true
 
 	lines := ll.getMarkedLines()
-	regex := regexp.MustCompile(`\[id "(\d+)"\]|"id":\s*"?(\d+)"?`)
+	regex := regexp.MustCompile(`\[id \\?"(\d+)\\?"\]|"id":\s*"?(\d+)"?`)
 	for _, line := range lines {
 		log.Trace().Msgf("ftw/waflog: Looking for any rule in '%s'", line)
 		match := regex.FindAllSubmatch(line, -1)
--- a/waflog/read.go
+++ b/waflog/read.go
@@ -28,6 +28,14 @@ func (ll *FTWLogLines) TriggeredRules() []uint {
 	ll.triggeredRulesInitialized = true
 
 	lines := ll.getMarkedLines()
+
+	// This regex provides flexibility in parsing how the rule ID is logged.
+	// `\[id \\?"(\d+)\\?"\]` supports:
+	//	- [id "999999"]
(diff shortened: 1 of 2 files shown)


waflog/read.go
```diff
@@ -28,7 +28,7 @@ func (ll *FTWLogLines) TriggeredRules() []uint {
 	ll.triggeredRulesInitialized = true
 
 	lines := ll.getMarkedLines()
-	regex := regexp.MustCompile(`\[id "(\d+)"\]|"id":\s*"?(\d+)"?`)
+	regex := regexp.MustCompile(`\[id \\?"(\d+)\\?"\]|"id":\s*"?(\d+)"?`)
```
theseion, line 31: Can you add a comment to describe the three different formats we're trying to detect here? I think the regex is now too complex to understand without context.

theseion (changes requested):
Please rebase.

waflog/read.go
M4tteoP: I proposed a regex description with some format examples. Writing it I realized that we would still not support `[\"id\":\"999999\"]`, however, I’m not aware of any engine or connector that outputs this format. Should we add support for consistency, or wait until a real use case arises?

theseion: Let's support what we know exists.

theseion (approved):

theseion:
Perfect, thanks!

PR #413, opened 2024-12-09
fix: flaky results: same start and end marker

M4tteoP:
I'm investigating some flakiness and found that the same marker might be retrieved for both the starter and the final one. When this happens, the function responsible for retrieving all the marked lines can not locate the starter marker. As a result, it uses all the logs up to the beginning of the file and processes IDs from lines belonging to previous tests, leading to incorrect results. 
My idea is that it can happen when go-ftw is too fast and reads again the same line before the engine/connector manages to write/flush the new logs (which include the final marker) into the file.

From a visual perspective (_Note: The read is performed backward._):

**] Expected scenario:**
start marker: `8888-8888`
final marker: `9999-9999`
```
...
log id:10
8888-8888
log id:11
9999-9999
```
1) We look for the final marker
2) Se start collecting log lines
3) We stop when reaching the start marker
4) log id:10 line is not collected, being out of the markers.

**] Assuming the case that both the start marker and final marker have the same value (equal to the start marker):**
start marker: `8888-8888`
final marker: `8888-8888`
```
...
log id:10
8888-8888
log id:11
9999-9999
```
1) We look for the final marker, in that case, the line `8888-8888` matches
2) From there we start collecting logs
3) `log id:11` is not collected, but rather `log id:10` and all the previous ones are collected.

The PR proposes to:
1) Add a debug log when the start marker is not found and unbounded logs (up to the beginning of the file) are used
2) Introduce a retry mechanism that, when looking for the final marker, repeats the seeking of the log file if the final marker found is equal to the start maker.
3) Do not introduce changes so far if the final marker is found equal to the start marker after some attempts, but print a verbose Warn log line.

It would allow us to have more visibility on the issue and bring in some more resiliency. Anyways I think that it is not expected to have the same start and final marker line, therefore I think adding a check would be beneficial.

I will follow with a test/reproducer whenever I manage to reproduce it in an easy and consistent way

### DIFF
--- a/check/base.go
+++ b/check/base.go
@@ -116,6 +116,10 @@ func (c *FTWCheck) GetTriggeredRules() []uint {
 	if c.CloudMode() {
 		return nil
 	}
+	// When a test is expecting to trigger an error and it effectively does, markers are not set.
+	if len(c.log.StartMarker()) == 0 || len(c.log.EndMarker()) == 0 {
+		return nil
+	}
 	return c.log.TriggeredRules()
 }
 
--- a/runner/run.go
+++ b/runner/run.go
@@ -165,6 +165,7 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 		Port:     testInput.GetPort(),
 		Protocol: testInput.GetProtocol(),
 	}
+
 	var startMarker []byte
 	if notRunningInCloudMode(ftwCheck) {
 		var err error
--- a/runner/run.go
+++ b/runner/run.go
@@ -165,9 +165,10 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 		Port:     testInput.GetPort(),
 		Protocol: testInput.GetProtocol(),
 	}
-
+	var startMarker []byte
 	if notRunningInCloudMode(ftwCheck) {
-		startMarker, err := markAndFlush(runContext, &testInput, stageId)
+		var err error
+		startMarker, err = markAndFlush(runContext, &testInput, stageId, nil)
 		if err != nil && !expectErr {
 			return fmt.Errorf("failed to find start marker: %w", err)
 		}
@@ -194,7 +195,7 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 	}
 
 	if notRunningInCloudMode(ftwCheck) {
-		endMarker, err := markAndFlush(runContext, &testInput, stageId)
+		endMarker, err := markAndFlush(runContext, &testInput, stageId, startMarker)
 		if err != nil && !expectErr {
 			return fmt.Errorf("failed to find end marker: %w", err)
 
@@ -221,7 +222,7 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 	return nil
 }
 
-func markAndFlush(runContext *TestRunContext, testInput *test.Input, stageId string) ([]byte, error) {
+func markAndFlush(runContext *TestRunContext, testInput *test.Input, stageId string, startMarker []byte) ([]byte, error) {
 	req := buildMarkerRequest(runContext, testInput, stageId)
 	dest := &ftwhttp.Destination{
 		DestAddr: testInput.GetDestAddr(),
@@ -239,7 +240,7 @@ func markAndFlush(runContext *TestRunContext, testInput *test.Input, stageId str
 			return nil, fmt.Errorf("ftw/run: failed sending request to %+v: %w", dest, err)
 		}
 
-		marker := runContext.LogLines.CheckLogForMarker(stageId, runContext.Config.MaxMarkerLogLines)
+		marker := runContext.LogLines.CheckLogForMarker(stageId, runContext.Config.MaxMarkerLogLines, startMarker)
 		if marker != nil {
 			return marker, nil
 		}
--- a/runner/run.go
+++ b/runner/run.go
@@ -24,8 +24,8 @@ import (
 )
 
 const (
-	// start and end UUID suffixes are used to disambiguate start and end markers.
-	// It permits to make the markers unique, while still maintaining one UUID per stage.
+	// Start and end UUID suffixes are used to disambiguate start and end markers.
+	// The suffixes make the markers unique, while still maintaining one UUID per stage.
 	startUuidSuffix = "-s"
 	endUuidSuffix   = "-e"
 )
--- a/runner/run.go
+++ b/runner/run.go
@@ -26,8 +26,8 @@ import (
 const (
 	// start and end UUID suffixes are used to disambiguate start and end markers.
 	// It permits to make the markers unique, while still maintaining one UUID per stage.
-	startUUIDSuffix = "-s"
-	endUUIDSuffix   = "-e"
+	startUuidSuffix = "-s"
+	endUuidSuffix   = "-e"
 )
 
 // Run runs your tests with the specified Config.
@@ -174,8 +174,8 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 	}
 
 	if notRunningInCloudMode(ftwCheck) {
-		startUUID := stageId + startUUIDSuffix
-		startMarker, err := markAndFlush(runContext, &testInput, startUUID)
+		startId := stageId + startUuidSuffix
+		startMarker, err := markAndFlush(runContext, &testInput, startId)
 		if err != nil && !expectErr {
 			return fmt.Errorf("failed to find start marker: %w", err)
 		}
@@ -202,8 +202,8 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 	}
 
 	if notRunningInCloudMode(ftwCheck) {
-		endUUID := stageId + endUUIDSuffix
-		endMarker, err := markAndFlush(runContext, &testInput, endUUID)
+		endId := stageId + endUuidSuffix
+		endMarker, err := markAndFlush(runContext, &testInput, endId)
 		if err != nil && !expectErr {
 			return fmt.Errorf("failed to find end marker: %w", err)
 
--- a/runner/run.go
+++ b/runner/run.go
@@ -23,6 +23,13 @@ import (
 	"github.com/coreruleset/go-ftw/waflog"
 )
 
+const (
+	// start and end UUID suffixes are used to disambiguate start and end markers.
+	// It permits to make the markers unique, while still maintaining one UUID per stage.
+	startUUIDSuffix = "-s"
+	endUUIDSuffix   = "-e"
+)
+
 // Run runs your tests with the specified Config.
 func Run(cfg *config.FTWConfiguration, tests []*test.FTWTest, c *RunnerConfig, out *output.Output) (*TestRunContext, error) {
 	out.Println("%s", out.Message("** Running go-ftw!"))
@@ -166,10 +173,9 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 		Protocol: testInput.GetProtocol(),
 	}
 
-	var startMarker []byte
 	if notRunningInCloudMode(ftwCheck) {
-		var err error
-		startMarker, err = markAndFlush(runContext, &testInput, stageId, nil)
+		startUUID := stageId + startUUIDSuffix
+		startMarker, err := markAndFlush(runContext, &testInput, startUUID)
 		if err != nil && !expectErr {
 			return fmt.Errorf("failed to find start marker: %w", err)
 		}
@@ -196,7 +202,8 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 	}
 
 	if notRunningInCloudMode(ftwCheck) {
-		endMarker, err := markAndFlush(runContext, &testInput, stageId, startMarker)
+		endUUID := stageId + endUUIDSuffix
+		endMarker, err := markAndFlush(runContext, &testInput, endUUID)
 		if err != nil && !expectErr {
 			return fmt.Errorf("failed to find end marker: %w", err)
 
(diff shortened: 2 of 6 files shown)


theseion:
What you want already exists:
1. `readLimit` limits the amount of backwards scanning to avoid reading the entire log file if it is large
2. if the start marker can't be found, another request is sent to try and flush the log and then the search is repeated, up to `maxMarkerRetries`

`CheckLogForMarker` returns the found line in each case and this line is then used as start / end marker. Even if they are identical, the search logic in `getMarkedLines` should ensure that finding the lines between the markers works.

I see one possible issue: when the log is not flushed immediately, `markAndFlush` will produce multiple marker lines in the log. If the start marker line is identical to these (e.g., there's no differing timestamp) then the these lines would match the start marker line and there wouldn't exist any other log lines in between. Example:

```
<marker ID>
log 1
log 2
<marker ID>
<marker ID>
```

We can solve this problem by using a suffix to the UUID (I'd like to keep a single UUID to make it easier to identify the matching markers when reading the log manually). E.g., the start marker could be `<UUID>-s` and the end marker `<UUID>-e`.

M4tteoP:
Thanks Max!

> 1. readLimit limits the amount of backwards scanning to avoid reading the entire log file if it is large
 
 In this scenario `readLimit` would just reduce the amount of unbounded log lines read, but, in the first place, we are reading the wrong log lines. So, with `readLimit` we would just read fewer wrong lines.

 > 2. if the start marker can't be found, another request is sent to try and flush the log and then the search is repeated, up to maxMarkerRetries.
 
 This logic does no help in this scenario: the start marker is correctly found, so no retries are performed.
 
 
> CheckLogForMarker returns the found line in each case and this line is then used as start / end marker. Even if they are identical, the search logic in getMarkedLines should ensure that finding the lines between the markers works.
 
If I'm getting the code right, `getMarkedLines` search logic is not working properly with two identical markers. So we should:

- Never give to `getMarkedLines` two identical markers.
- Or adapt `getMarkedLines` to not really look for the identical markers, but just rely on the UUID (but then there is no point on passing to that function the two marker lines because it will have to find them on its own).

 
Going step by step:
1. `markAndFlush` is called, a request is performed to log the starter marker. Now logs will be like this:
```
...
log 0   <-- old log belonging to the previous request
[msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "WWAVDNpnKkXiHleWcdD"]
```
2. `markAndFlush` internally calls `CheckLogForMarker` which looks, starting from the end, for a line with `7d16c799-954f-4338-81c9-fb39b32a9521`. The line is found and identified as the start marker line. No retries are performed.
3. `RunStage` stores the start marker via `SetStartMarker`. The start marker now is `[msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "WWAVDNpnKkXiHleWcdD"]`.
4. `RunStage` performs the real request that might or might not trigger some rules and the related logs will be printed.
5. `markAndFlush` is called again to print the end marker. The request is performed correctly, but the Engine is not fast enough on writing rules logs and the end marker.
6. `markAndFlush` internally calls `CheckLogForMarker` when logs are looking like the following:
```
...
log 0 
[msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "WWAVDNpnKkXiHleWcdD"]
log 1
log 2
```
We can notably see that we don't have yet a trailing marker with `7d16c799-954f-4338-81c9-fb39b32a9521` but a different unique_id.

7. `CheckLogForMarker` looks again for a line with `7d16c799-954f-4338-81c9-fb39b32a9521` and within `readLimit` lines, it finds `[msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "WWAVDNpnKkXiHleWcdD"]`
8. `RunStage` stores the final marker via `SetEndMarker` using the same line of the start marker.
Start marker: `[msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "WWAVDNpnKkXiHleWcdD"]`
End marker: `[msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "WWAVDNpnKkXiHleWcdD"]`

9. At this point `getMarkedLines` is called, providing start and end marker.
10. (not so relevant, this point might not happen and the outcome would be the same) In the meanwhile the Engine might managed to write the whole logs. So they might now look like the following:
```
...
log 0 
[msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "WWAVDNpnKkXiHleWcdD"]
log 1
log 2
log 3
[msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "gfrFHrMGygNvrCkkWRY"]
```
11. `getMarkedLines` starts from the bottom and looks for the end marker. The real end marker (`gfrFHrMGygNvrCkkWRY`) and all the logs related to this request are skipped. It finally finds the end marker.
```
...
log 0 
⬆ [msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "WWAVDNpnKkXiHleWcdD"]
❌ log 1
❌ log 2
❌ log 3
❌ [msg "X-CRS-Test 7d16c799-954f-4338-81c9-fb39b32a9521"][unique_id "gfrFHrMGygNvrCkkWRY"]
```
12. From `WWAVDNpnKkXiHleWcdD` it starts reading old log lines (such as `log 0`) until:
- it finds the start marker: but it will never happend, because the start marker has been misused as end marker
- or it reaches `readLimit` reading `readLimit` wrong log lines
- or it reaches the beginning of the document.
13. All these log lines are then used to extract IDs and computing wrong outcomes. The correct logs to be used (`log 1`, `log 2`, `log 3`) are not taken into account

theseion:
Thanks for the analysis. You're right, in that case both markers would be the same line. Both cases, the one you described and the one I came up with, would be fixed by using identifying suffixes for the start and end markers, since `CheckLogForMarker` could no longer confuse the two.

M4tteoP:
Thank you! It sounds like a very reasonable idea to have the same `uuid` just followed by a suffix.
I have:
- Implemented the suffix idea
- Removed the SeekAttempts
- Kept some more debug verbosity when markers are not found
- Added some logic to not consider as marked lines all the lines found before finding the end marker (backward read). The tests were adding one or more empty lines, in real-case scenarios I guess there could be logs that we should just ignore

runner/run.go
```diff
@@ -166,7 +173,8 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 	}
 
 	if notRunningInCloudMode(ftwCheck) {
-		startMarker, err := markAndFlush(runContext, &testInput, stageId)
+		startUUID := stageId + startUUIDSuffix
```
theseion, line 176:
```suggestion
		startId := stageId + startUUIDSuffix
```

runner/run.go
```diff
@@ -22,6 +22,13 @@ import (
 	"github.com/coreruleset/go-ftw/waflog"
 )
 
+const (
+	// start and end UUID suffixes are used to disambiguate start and end markers.
+	// It permits to make the markers unique, while still maintaining one UUID per stage.
+	startUUIDSuffix = "-s"
+	endUUIDSuffix   = "-e"
```
theseion, line 29:
```suggestion
	startUuidSuffix = "-s"
	endUuidSuffix   = "-e"
```

runner/run.go
```diff
@@ -190,7 +198,8 @@ func RunStage(runContext *TestRunContext, ftwCheck *check.FTWCheck, testCase sch
 	}
 
 	if notRunningInCloudMode(ftwCheck) {
-		endMarker, err := markAndFlush(runContext, &testInput, stageId)
+		endUUID := stageId + endUUIDSuffix
```
theseion, line 201:
```suggestion
		endId := stageId + endUUIDSuffix
```

waflog/read.go
```diff
@@ -158,24 +163,34 @@ func (ll *FTWLogLines) getMarkedLines() [][]byte {
 			break
 		}
 		lineLower := bytes.ToLower(line)
-		if !endFound && bytes.Equal(lineLower, ll.endMarker) {
-			endFound = true
+
+		if !endFound {
+			// Skip lines until we find the end marker. Reading backwards, the line we are looking for are
```
theseion, line 168:
```suggestion
			// Skip lines until we find the end marker. Reading backwards, the lines we are looking for are
```

waflog/read.go
```diff
@@ -158,24 +163,34 @@ func (ll *FTWLogLines) getMarkedLines() [][]byte {
...
+			startFound = true
 			break
 		}
 
 		saneCopy := make([]byte, len(line))
 		copy(saneCopy, line)
 		ll.markedLines = append(ll.markedLines, saneCopy)
 	}
+	if !startFound {
+		log.Debug().Msgf("starter marker not found while getting marked lines")
```
theseion, line 185:
```suggestion
		log.Debug().Msg("starter marker not found while getting marked lines")
```

waflog/read.go
```diff
@@ -158,24 +163,34 @@ func (ll *FTWLogLines) getMarkedLines() [][]byte {
...
+	}
 
 	return ll.markedLines
 }
 
 // CheckLogForMarker reads the log file and searches for a marker line.
 // stageID is the ID of the current stage, which is part of the marker line
+// At this point, stageID includes also the start or end suffix.
 // readLimit is the maximum numbers of lines to check
 func (ll *FTWLogLines) CheckLogForMarker(stageID string, readLimit uint) []byte {
```
theseion, line 195:
```suggestion
func (ll *FTWLogLines) CheckLogForMarker(markerId string, readLimit uint) []byte {
```

waflog/read.go
```diff
@@ -158,24 +163,34 @@ func (ll *FTWLogLines) getMarkedLines() [][]byte {
...
+	if !startFound {
+		log.Debug().Msgf("starter marker not found while getting marked lines")
+	}
 
 	return ll.markedLines
 }
 
 // CheckLogForMarker reads the log file and searches for a marker line.
 // stageID is the ID of the current stage, which is part of the marker line
+// At this point, stageID includes also the start or end suffix.
```
theseion, line 193:
```suggestion
// markerId is the ID of the current stage + suffix (for start / end), which is part of the marker line
```

waflog/read_test.go
```diff
@@ -201,12 +201,15 @@ func (s *readTestSuite) TestFTWLogLines_Contains() {
 	s.NotNil(cfg)
 
 	stageID := "dead-beaf-deadbeef-deadbeef-dead"
```
theseion, line 203:
```suggestion
	stageId := "dead-beaf-deadbeef-deadbeef-dead"
```

waflog/read_test.go
```diff
@@ -368,12 +373,13 @@ func (s *readTestSuite) TestFindAllIdsInLogs() {
...
```
theseion, line 375:
```suggestion
	stageId := "dead-beaf-deadbeef-deadbeef-dead"
```

theseion (changes requested):
We don't have a test yet, that specifically reproduces the issue, right? Could you add one?

fzipi:
@M4tteoP Can you address comments here and fix the conflict so we merge this one?

M4tteoP:
Yes, It is in my plans, I will be back on this one soon

M4tteoP:
It should finally be ready for another look. Apologies for the delay!

check/base.go
```diff
@@ -116,6 +116,10 @@ func (c *FTWCheck) GetTriggeredRules() []uint {
 	if c.CloudMode() {
 		return nil
 	}
+	// When a test is expecting to trigger an error and it effectively does, markers are not set.
```
M4tteoP, line 119: After https://github.com/coreruleset/go-ftw/pull/432, `GetTriggeredRules()` is now called not only when we are looking for IDs in the logs based on the test expectations but also at the end of the test to provide the list of triggered rules. When a test is expected to trigger an error, start and end markers might not populated. As a result, calling TriggeredRules() in such cases triggers a fatal error. Similar to how [we do not support looking for IDs when expecting an error](https://github.com/coreruleset/go-ftw/blob/528e92cae88cd7c084e087856214f71f687cc4dc/runner/run.go#L371-L374), I believe it is reasonable not to look for triggered rules in this case.

check/base.go
```diff
@@ -116,6 +116,10 @@ func (c *FTWCheck) GetTriggeredRules() []uint {
...
+	if len(c.log.StartMarker()) == 0 || len(c.log.EndMarker()) == 0 {
+		return nil
+	}
```
M4tteoP, line 122: After https://github.com/coreruleset/go-ftw/pull/432, `GetTriggeredRules()` is now called not only when we are looking for IDs in the logs based on the test expectations but also at the end of the test to provide the list of triggered rules. When a test is expected to trigger an error, start and end markers might not populated. As a result, calling TriggeredRules() in such cases triggers a fatal error. Similar to how [we do not support looking for IDs when expecting an error](https://github.com/coreruleset/go-ftw/blob/528e92cae88cd7c084e087856214f71f687cc4dc/runner/run.go#L371-L374), I believe it is reasonable not to look for triggered rules in this case.

theseion: Yes, I agree.

runner/run.go
```diff
@@ -23,6 +23,13 @@ import (
 	"github.com/coreruleset/go-ftw/waflog"
 )
 
+const (
+	// start and end UUID suffixes are used to disambiguate start and end markers.
```
theseion, line 27:
```suggestion
	// Start and end UUID suffixes are used to disambiguate start and end markers.
```

runner/run.go
```diff
@@ -23,6 +23,13 @@ import (
...
+	// It permits to make the markers unique, while still maintaining one UUID per stage.
```
theseion, line 28:
```suggestion
	// The suffixes make the markers unique, while still maintaining one UUID per stage.
```

waflog/read.go
```diff
@@ -165,29 +170,38 @@ func (ll *FTWLogLines) getMarkedLines() [][]byte {
...
+			startFound = true
 			break
 		}
 
 		saneCopy := make([]byte, len(line))
 		copy(saneCopy, line)
 		ll.markedLines = append(ll.markedLines, saneCopy)
 	}
+	if !startFound {
+		log.Debug().Msg("starter marker not found while getting marked lines")
```
theseion, line 192:
```suggestion
		log.Debug().Msg("start marker not found while collecting marked lines")
```

waflog/read_test.go
theseion, line 1: We don't have a test yet, that specifically reproduces the issue, right? Could you add one?

theseion (changes requested):

waflog/read_test.go
M4tteoP: - With your idea of splitting the markers, the original issue ends up being an end marker not found by `CheckLogForMarker`, I slightly changed the already existing `TestReadCheckLogForMarkerNoMarkerAtEnd`.
- I added a new test (`TestReadCheckLogForMarkerWithMultipleMarkersAtEnd`) for the other potential issue that you [highlighted](https://github.com/coreruleset/go-ftw/pull/413#issuecomment-2530585028) (

theseion (approved):

PR #420, opened 2024-12-14
fix: use correct names for flags

### LINKED ISSUE #419: wait-delay flag not used
We have a flag called `wait-delay` which has no effect...

theseion:
Fixes #419


fzipi (approved):
Perfect refactor! So much easy to read 😄

PR #423, opened 2024-12-19
chore: remove raw input references

### LINKED ISSUE #408: deprecate `raw` input in tests
We have been using and recommending `encoded` instead. We should deprecate `raw` input in tests and remove all the logic we have for that in go-ftw.

fzipi:
## what

- remove raw input references in code

## why

- cleanup

## references

- Fixes #408


theseion:
We should accompany this PR with a PR to the schema to remove raw data there as well and then use a new version of the schema in the next release.

Issue #428, opened 2024-12-29
Fields output.log_contains and input.encoded_data described in specs not working


Sebitosh:
Hello,

I am working on the ModSecurity Regression Test Set.
Found this issues while working on owasp-modsecurity/MRTS#6 & owasp-modsecurity/MRTS#12 . 

I noticed that `log_contains` and `no_log_contains` checks have no effect on test results while still being described in the [latest yaml schema specs](https://github.com/coreruleset/ftw-tests-schema/blob/main/spec/v2.1.1/ftw.md#output). It seems that fields `log.match_regex` and `log.no_match_regex` replaced them entirely ?

I also noticed that while `input.encoded_request` worked fine, `input.encoded_data` had no effect on the generated request used in a test case. It is also still described in the [specs](https://github.com/coreruleset/ftw-tests-schema/blob/main/spec/v2.1.1/ftw.md#input).

Here are the files to demonstrate the issue (there is yaml tests using the fields, the rules I used, a commented pcapng showing the requests send by go-ftw):  [issue-go-ftw-log-contains-encoded-data.zip](https://github.com/user-attachments/files/18269220/issue-go-ftw-log-contains-encoded-data.zip)

Output I get when running these tests:
~~~
🚀 Running go-ftw!
👉 executing tests in file DEMO_ENCODED_DATA.yaml
	running 100001-1: ✔ passed in 8.397749ms (RTT 52.272652ms)
	running 100001-2: 💥 100001-2 failed in 7.556905ms (RTT 51.998539ms)
	running 100001-3: 💥 100001-3 failed in 7.536512ms (RTT 52.249954ms)
👉 executing tests in file DEMO_LOG_CONTAINS.yaml
	running 100000-1: ✔ passed in 7.921806ms (RTT 52.490335ms)
	running 100000-2: 💥 100000-2 failed in 7.079077ms (RTT 51.914455ms)
➕ run 5 total tests in 38.492049ms
⏭ skipped 0 tests
👎 3 test(s) failed to run: ["100001-2" "100001-3" "100000-2"]
Error: failed 3 tests
sebitosh@lp:~$ go-ftw -v
ftw version 1.1.2
commit: 2c1ba3070af090cbdba6603a639245f8cb4b2108
built at: 2024-11-27T06:01:26Z
built by: goreleaser

~~~

theseion:
[...]

PR #431, opened 2024-12-30
feat: implement `encoded_data`

theseion:
`encoded_data` has been part of the spec for a while but never implemented. Allows passing body data base64 encoded for complex bodies.

Refs #428


fzipi (approved):
Awesome 😎

PR #432, opened 2024-12-30
feat: report triggered rules for every test

### LINKED ISSUE #429: YAML Configuration: Test Fails When Whitelisted Requests Trigger WAF Rule ID Alerts
Could you please explain how to configure YAML so that if none of the alert rules are satisfied, we can collect certain whitelisted requests. These whitelisted requests should not match any alerts. After configuring the YAML test file, if any WAF rule ID alerts are triggered, it should be considered a test failure, indicating that the rules need to be optimized.

```
rule_id: 932140
tests:
  - test_id: 1
    desc: for %
    stages:
      - input:
          dest_addr: 127.0.0.1
          headers:
            Accept: "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5"
            Host: localhost
            User-Agent: "OWASP CRS test agent"
          method: GET
          port: 80
          uri: "/get?foo=for%20%25variable%20in%20%28set%29%20do%20command"
          version: HTTP/1.0
        output:
          log:
            expect_ids: [932140]
```

no rule_id expect_ids，just test case id

theseion:
Adds list of triggered rules by stage to the JSON output.

Fixes #429

### DIFF
--- a/check/base.go
+++ b/check/base.go
@@ -112,6 +112,13 @@ func (c *FTWCheck) SetEndMarker(marker []byte) {
 	c.log.WithEndMarker(marker)
 }
 
+func (c *FTWCheck) GetTriggeredRules() []uint {
+	if c.CloudMode() {
+		return nil
+	}
+	return c.log.TriggeredRules()
+}
+
 func (c *FTWCheck) Close() error {
 	return c.log.Cleanup()
 }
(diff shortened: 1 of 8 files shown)


fzipi (approved):

theseion:
Requires #436.

theseion:
ping @fzipi

fzipi (approved):
LGTM

PR #434, opened 2025-01-03
fix: output ftw result error

xu-xiang:
```
 --output json --file r.json
```

output test result eroor with 
```
Error: open r.json: no such file or directory
```
change Open to  Create


theseion (approved):

theseion:
Thanks! Please consider setting up commit signing. We usually allow only signed commits in our repositories.

PR #443, opened 2025-01-12
feat: implement multi header support

### LINKED ISSUE #332: Implement multi-value headers
The `Header` type currently doesn't support multiple values for a single header. We need to support requests that include the same header multiple times, possibly with different casings. Examples:

```
// different casing
Content-Type: application/json
ConTent-tyPE: application/json
// same casing, same or different value
content-type: application/json
content-type: application/json
content-type: application/problem+json
```

**Acceptance criteria**
- it must be possible to specify headers as in the examples above in a test case (YAML)
- it must be possible to add multiple values to a specific header name
- retrieving headers must return a map, as a key can map multiple versions of a header

theseion:
- support multiple headers with the same name
- support multiple headers with identical name and value
- make list of headers ordered; in the future, tests should be able to enforce the order of headers in a request
- improve API, tests, and documentation of ftwhttp.Header

- disable logging in tests where possible
- enable self-updater test (go-critic was complaining because the test file was touched)

Fixes #332


theseion:
Note that our test format currently does not support multiple headers with the same name, because we use a map for headers. In the future we should use a list instead in order to allow for repeating headers and values.

ftwhttp/header.go
```diff
@@ -4,149 +4,182 @@
...
-// The key is case-insensitive
-func (h Header) Get(key string) string {
-	if h == nil {
-		return ""
+// Creates a new Header from a map of HTTP header names and values.
+//
+// This is a convenience and legacy fallback method. In the future,
+// headers should be specified as a list, in order to guarantee order
+// and to allow requests to contain the same header multiple times,
+// potentially, but not necessarily, with different values.
```
fzipi, line 48:
```suggestion
// potentially with different values.
```
I think that _potentially_ already offers the needed conditional future.

ftwhttp/header.go
```diff
@@ -4,149 +4,182 @@
...
 
-// Clone returns a copy of h
-func (h Header) Clone() Header {
-	clone := make(Header)
+// Write writes the header to the provided writer
+func (h *Header) Write(writer io.Writer) error {
+	buf := bufio.NewWriter(writer)
+	for _, tuple := range h.entries {
+		if log.Trace().Enabled() {
+			log.Trace().Msgf("Writing header: %s: %s", tuple.Name, tuple.Value)
```
fzipi, line 126: Very subtle, but as headers can have the same name and value, could it be that the first time it doesn't fail, but the second one does? Meaning: do we want to use the discarded `for _, tuple := range` and print it? E.g. `for i, tuple := range.... log.Trace().Msgf(... " i )`

ftwhttp/header.go
```diff
@@ -4,149 +4,182 @@
...
+		}
+		if _, err := buf.WriteString(headerSeparator); err != nil {
+			return err
+		}
+		if _, err := buf.WriteString(tuple.Value); err != nil {
+			return err
+		}
+		if _, err := buf.WriteString(headerDelimiter); err != nil {
+			return err
+		}
```
fzipi, line 139: Does it add value doing it like this? Why not all four at once, and return the error?

ftwhttp/header_names/constants.go
fzipi, line 1: Hmmm.. do we need a separate package for this? Why not adding constants in `ftwhttp/header.go` with a prefix like `HeaderConnection` or `HeaderConnectionName`?

ftwhttp/header_test.go
```diff
@@ -1,64 +1,50 @@
-// Copyright 2010 The Go Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style
-// license that can be found in the LICENSE file.
-
-// Copyright 2011 The Go Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style
-// license that can be found in the LICENSE file.
-
```
fzipi, line 8: 🤦

ftwhttp/header_values/constants.go
```diff
@@ -0,0 +1,5 @@
+package ftwhttp
+
+const (
+	ApplicationXWwwFormUrlencoded = "application/x-www-form-urlencoded"
+)
```
fzipi, line 5: Ok. Maybe creating a `header` package instead? And moving all constants there?

ftwhttp/header_values/constants.go
```diff
@@ -0,0 +1,5 @@
+package ftwhttp
```
fzipi, line 1:
```suggestion
package header_values
```

ftwhttp/request.go
```diff
@@ -127,7 +134,7 @@ func BuildRequest(r *Request) ([]byte, error) {
 	}
 
 	// Multipart form data needs to end in \r\n, per RFC (and modsecurity make a scene if not)
-	if ct := r.headers.Value(ContentTypeHeader); strings.HasPrefix(ct, "multipart/form-data;") {
+	if r.headers.HasAnyValueContaining(header_names.ContentType, "multipart/form-data;") {
 		crlf := []byte("\r\n")
```
fzipi, line 138: We have a constant for headerSeparator, maybe we want to use it here? https://github.com/coreruleset/go-ftw/pull/443/files#diff-fad61aedc5779df83ed3dfea5f520c19e259eafc9e86bdf09447d7dab9706e65R18

ftwhttp/request.go
```diff
@@ -140,28 +147,21 @@ func BuildRequest(r *Request) ([]byte, error) {
...
-	// TODO: handle cookies
-	// if client.Jar != nil {
-	// 	for _, cookie := range client.Jar.Cookies(req.URL) {
-	// 		req.AddCookie(cookie)
-	// 	}
-	// }
-
 	// After headers, we need one blank line
-	_, err = fmt.Fprintf(&b, "\r\n")
+	_, err = b.WriteString("\r\n")
```
fzipi, line 157: Same here? We have a constant for headerSeparator, maybe we want to use it here? https://github.com/coreruleset/go-ftw/pull/443/files#diff-fad61aedc5779df83ed3dfea5f520c19e259eafc9e86bdf09447d7dab9706e65R18

fzipi (changes requested):
Thanks for taking this. Big change.

I'll suggest you to split in 3 PRs, as we are mixing stuff here.

1. All the zerolog changes (🙏 good catch). E.g. inclusion and initialization.
```
func (s *responseTestSuite) SetupSuite() {
	zerolog.SetGlobalLevel(zerolog.Disabled)
}
```
2. The uncommented `self_updater` tests
3. This PR, rebased after the above.

It will make things more clear if we need to revert something.

ftwhttp/header.go
theseion: Done

ftwhttp/header.go
theseion: This is a tiny bit faster, according to the benchmark. Probably not worth it though, you're right.

ftwhttp/header_values/constants.go
theseion: My problem in Go is namespacing. If both names and values are in the same package, then we need to add some prefix to every entry to make it clear what the constant is. Using distinct packages solves this namespacing problem, even though it's pretty ugly.

ftwhttp/header_names/constants.go
theseion: See my other comment.

ftwhttp/header_values/constants.go
fzipi: Yeah. We cannot have both. Or we have the prefixed version, or we have ugly namespaces.

theseion: I asked @M4tteoP for his opinion on this and he agreed that the way it is now it's at least readable. He didn't see a better option either. Should we leave it this way?

theseion: Ping @fzipi.

test/data_test.go
```diff
@@ -33,13 +34,13 @@ port: 80
 headers:
 User-Agent: "ModSecurity CRS 3 Tests"
 Host: "localhost"
-Content-Type: "application/x-www-form-urlencoded"
+Content-Type: header_values.ApplicationXWwwFormUrlencoded
```
fzipi, line 37: Does this require a string substitution? I mean, is this part of the template? E.g.
```suggestion
Content-Type: {{ header_values.ApplicationXWwwFormUrlencoded }}
```
?

test/data_test.go
```diff
@@ -53,14 +54,14 @@ port: 80
...
```
fzipi, line 57: Same here.

test/data_test.go
```diff
@@ -75,19 +76,20 @@ port: 80
...
```
fzipi, line 79: And here.

test/data_test.go
```diff
@@ -101,19 +103,20 @@ port: 80
...
```
fzipi, line 106: ++

test/data_test.go
```diff
@@ -127,19 +130,20 @@ port: 80
...
```
fzipi, line 133: ++

test/data_test.go
```diff
@@ -151,18 +155,19 @@ port: 80
...
```
fzipi, line 158: ++

test/data_test.go
```diff
@@ -174,17 +179,18 @@ port: 80
...
```
fzipi, line 182: ++

If we use it so many times, do we need to repeat it this much? 🤔

theseion: The strings have small differences, so it's not so easy to extract a template without parameterizing everything.

fzipi (dismissed):

theseion:
Extended PR with updated schema and support for new `ordered_headers` field.

theseion:
Please review @fzipi @M4tteoP

fzipi (approved):
LGTM. I see a missed opportunity for _pineapple_ and _pizza_ vs. _homer_ & _donuts_ 😄 😄

PR #457, opened 2025-02-14
feat: support global overriding of virtual_host_mode

### LINKED ISSUE #453: Cannot set host header
Experimentally was not possible to set the `Host` header, neither via config file nor via yaml test files, was always using `Host: localhost`.

- `ftw_1.2.0_amd64.deb` (on Ubuntu 22.04)
- Settings `dest_addr` and `port` in config file did work, i.e. was connecting to the correct host+port, but with HTTP requests with `Host: localhost`
- (Setting `headers` > `Host` in config file had no effect, also not setting in yaml test files)
- Made no difference if config file was at a default location or indicated in command line with `--config`
- Trace output of ftw (`--trace`) did always show `Host: localhost`
- (Could not find any indication for the observed behavior in the source code, would seem to merge headers as documented)

(As a workaround in our case, because the Envoy instance to test was in Kubernetes, we configured the ingress to send requests to `localhost` to the Envoy instance to test, i.e. no urgency of any sort on our side...)

theseion:
fixes #453


theseion:
@airween with this PR you can use the global configuration to set `virtual_host_mode` instead of having to patch test files.

fzipi (approved):

PR #474, opened 2025-03-15
feat: parameterize test file matching

theseion:
- expose `--glob` / `-g` to allow users to override the default filename glob
- improve default filename glob to match both `*.yaml` and `*.yml`

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -25,6 +25,7 @@ import (
 const (
 	connectTimeoutFlag               = "connect-timeout"
 	dirFlag                          = "dir"
+	globFlag                         = "glob"
 	excludeFlag                      = "exclude"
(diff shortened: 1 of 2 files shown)


cmd/run.go
```diff
@@ -103,13 +105,10 @@ func runE(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	dir, _ := cmd.Flags().GetString(dirFlag)
-	files := fmt.Sprintf("%s/**/*.yaml", dir)
-	tests, err := test.GetTestsFromFiles(files)
+	tests, err := loadTests(cmd)
```
fzipi, line 108: ❤️

fzipi (approved):

PR #504, opened 2025-06-05
chore: refactor runner configuration

theseion:
The `check` package is only used from the `runner` package. It makes sense to use `TestRunContext` and other types in `check` but that is not possible due to circular imports. Make `check` part of `runner` instead.

In addition, to reduce issues with cyclic imports and confusion between different configuration objects, the `RunnerConfig` now consolidates all configuration data for the test runner.


ftwhttp/client.go
```diff
@@ -7,12 +7,13 @@ import (
...
 	"crypto/tls"
 	"crypto/x509"
-	"fmt"
 	"net"
 	"net/http/cookiejar"
 	"strings"
 	"time"
 
+	"fmt"
+
```
fzipi, line 16: Why does it need to be separated?

fzipi (dismissed):
That was a long one. Some nits.

config/config.go
```diff
@@ -100,80 +96,35 @@ func NewConfigFromString(conf string) (*FTWConfiguration, error) {
...
-}
-
-// WithMaxMarkerLogLines sets the new amount of lines we go back in the logfile attempting to find markers.
-func (c *FTWConfiguration) WithMaxMarkerLogLines(amount uint) {
-	c.MaxMarkerLogLines = amount
-}
+// // WithMaxMarkerLogLines sets the new amount of lines we go back in the logfile attempting to find markers.
+// func (c *FTWConfiguration) WithMaxMarkerLogLines(amount uint) {
+// 	c.MaxMarkerLogLines = amount
+// }
```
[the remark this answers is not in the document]
theseion: Thanks. That was left over.

ftwhttp/client.go
theseion: Not on purpose.

fzipi (dismissed):
Thanks!

fzipi (approved):

PR #631, opened 2026-05-09
fix: skip duplicate end markers when collecting marked lines

### LINKED ISSUE #625: sometimes 999999-XX tests randomly fail (coraza)
*i think this is new since upgraded to CRS 4.25 (or coraza 3.7.0), at least i never had this issue before*

sometimes when we run the complete test suite, one of the 999999-XX tests fails. 
* It does not always happen, and it is not always the same test. 
* It happens not very often, maybe once per 20 test runs, or even less

for example here it is 999999-31
https://github.com/united-security-providers/coraza-envoy-go-filter/actions/runs/24498005874/job/71597446754?pr=91#step:6:6187

and here it is 999999-20
https://github.com/united-security-providers/coraza-envoy-go-filter/actions/runs/24457058007/job/71460515736#step:6:6162

theseion:
When forcing the log to flush with `markAndFlush()`, it might contend multiple end markers, as they will be resent until found in the log. The "marked lines" are expected to contain only the lines _between_ the start and end marker, and any end marker in there might lead to false test results accidentally.

Fixes #625


fzipi (approved):
Awesome

PR #656, opened 2026-07-22
feat(quantitative): add --threshold and multi-rule --rule support

### LINKED ISSUE #387: feature: add threshold for quantitative testing
## Requirements

- add a `--threshold` or similar to when passed as a variable, it will exit with an error code
- it should be supported when testing only one rule also e.g. `-r <rule> -t <threshold>` should fail only when that specific rule is above the threshold
- as we use a ratio, the threshold should be a _float_ probably

fzipi:
## Summary

- `ftw quantitative` gains a `-t`/`--threshold` flag: fails with a non-zero exit code when a requested rule's false-positive ratio exceeds the given value, without needing `--baseline`/`--compare-crs`.
- `-r`/`--rule` now accepts multiple rule IDs (repeated flag or comma-separated), so a single CI run can check a threshold across every rule affected by a shared regex-assembly change.
- `--threshold` requires `--rule` to be set and rejects negative values.
- The threshold check result (per rule: ratio, pass/fail) is included in plain-text, JSON (`thresholdCheck` field), and Markdown output.

Closes #387.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `golangci-lint run`, `gofumpt`/`goimports` all clean
- [x] Unit tests for `evaluateThreshold`/`thresholdError` covering pass/fail/disabled/unknown-rule cases
- [x] Unit tests for `FalsePositiveRatioForRule` and threshold reporting in plain/JSON/Markdown output
- [x] Command-level tests for flag parsing/validation and multi-rule support
- [x] End-to-end command-level JSON test asserting the `thresholdCheck` field round-trips
- [x] `go test ./...` passes for touched packages

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added support for selecting multiple rule IDs with repeated or comma-separated `--rule`/`-r` options.
  - Added `--threshold`/`-t` to set an acceptable false-positive ratio per selected rule.
  - Threshold check results are now included in text, Markdown, and JSON output with per-rule pass/fail details.
  - Quantitative runs now report an error when the configured threshold is exceeded.
- **Bug Fixes**
  - Added validation to reject negative `--threshold` values and to prevent using `--threshold` without specifying at least one `--rule`.


cmd/quantitative/quantitative.go
```diff
@@ -64,7 +65,8 @@ func New(cmdContext *internal.CommandContext) *cobra.Command {
 	runCmd.Flags().Bool(allParanoiaLevelsFlag, false, "Evaluate all CRS paranoia levels in one run.")
 	runCmd.Flags().IntP(corpusLineFlag, "n", 0, "Number is the payload line from the corpus to exclusively send.")
 	runCmd.Flags().StringP(payloadFlag, "p", "", "Payload is a string you want to test using quantitative tests. Will not use the corpus.")
-	runCmd.Flags().IntP(ruleFlag, "r", 0, "Rule ID of interest: only show false positives for specified rule ID. Defaults to paranoia level 4 unless -P is also set.")
+	runCmd.Flags().IntSliceP(ruleFlag, "r", nil, "Rule ID(s) of interest: only show false positives for the specified rule ID(s). Can be repeated or comma-separated, e.g. -r 920100 -r 920120. Defaults to paranoia level 4 unless -P is also set.")
+	runCmd.Flags().Float64P(thresholdFlag, "t", 0, "Maximum acceptable false-positive ratio for each rule given via --rule; exceeding it for any rule causes a non-zero exit code. Requires --rule.")
```
theseion: Improve explanation. As a user, I have no idea what that "ratio" will look like. Do I have to specify `0.12`? or `1200`?

theseion (changes requested):

cmd/quantitative/quantitative.go
fzipi: Updated with a ratio example.

theseion (approved):

PR #664, opened 2026-09-05
fix(output): emit valid GitHub Actions workflow commands for -o github

### LINKED ISSUE #663: -o github output emits malformed GitHub workflow commands
## Bug

The `github` output type (`ftw run -o github`) does not produce valid
GitHub Actions workflow commands. In `output/output.go`, the `GitHub`
case of `Printf` does:

```go
s = fmt.Sprintf(format, a...)
s = fmt.Sprintf("::notice file={name},line={line},endLine={endLine},title={title}::{%s}", s)
```

Two problems:

1. `{name}`, `{line}`, `{endLine}` and `{title}` are literal text — they
   are never substituted with real values, so every line emits an
   annotation pointing at a non-existent file literally named `{name}`.
2. The message is wrapped in braces (`::{%s}`), which is not part of the
   workflow command syntax.

GitHub Actions ignores malformed commands, so `-o github` renders no
annotations at all — worse than the plain output, since the log only
contains noise.

## Expected behavior

Valid commands per the workflow command syntax, e.g.:

```
::notice::- 999999-1 failed in 149ms (RTT 85ms)
::error::- 920100-1 failed in 150ms (RTT 87ms)
```

with the message body URL-escaped (`%` → `%25`, `\r` → `%0D`,
`\n` → `%0A`), one command per line, and failed tests reported with
`::error::` severity. `file`/`line`/`endLine` should be omitted rather
than emitted as unsubstituted placeholders.

Happy to open a PR with the fix.

A13501350:
## What

Fix the `github` output type so it emits valid GitHub Actions workflow
commands instead of malformed ones.

Before (broken — unsubstituted placeholders, stray braces, GitHub drops
the command):

```
::notice file={name},line={line},endLine={endLine},title={title}::{+ passed in 148ms (RTT 84ms)}
```

After:

```
::notice::+ passed in 148ms (RTT 84ms)
::error::- 999999-1 failed in 149ms (RTT 85ms)
```

Changes in `output/output.go`:

- replace the broken wrapper with a `githubCommand` builder emitting
  `::severity::message`
- escape the message body per the actions runner (`%` → `%25`,
  `\r` → `%0D`, `\n` → `%0A`); keep `Println`'s line break as a real
  newline so each command sits on its own line
- add an `AnnotationSeverity` type (`notice`/`warning`/`error`) with a
  `SetSeverity` setter

Changes in `runner/run.go`:

- `displayResult` reports failed tests as `::error::` and resets to
  `notice` afterwards

`NewOutput` uses the plain catalog for github output so `:emoji:` codes
from the normal catalog do not leak into annotation text.

## Why

`-o github` is meant to surface test results as GitHub annotations. The
current output is rejected by Actions, so the feature silently produces
nothing.

Fixes #663

## Testing

- `output`/`runner` unit tests, including new cases for the command
  format, escaping (`%`, CR, LF), error severity and the real-newline
  handling in `Println`
- validated end-to-end in a real Actions run of the CRS regression suite
  (ModSecurity IIS connector, ~4900 tests): failures rendered as error
  annotations, summary lines as notices

## Summary by CodeRabbit

* **New Features**
  * GitHub Actions annotations now support notice, warning, and error severity levels.
  * Failed results are reported as error annotations, while other output remains notice-level by default.

* **Bug Fixes**
  * Annotation messages now safely handle special characters, including percent signs and line breaks.
  * Multiple annotations remain separately parseable when printed together.
  * Error annotations no longer include unnecessary file and line details.

### DIFF
--- a/output/output.go
+++ b/output/output.go
@@ -26,6 +26,15 @@ const (
 	Markdown Type = "markdown" // markdown-friendly plain text
 )
 
+// AnnotationSeverity is the severity of a GitHub Actions workflow command.
+type AnnotationSeverity string
+
+const (
+	AnnotationNotice  AnnotationSeverity = "notice"
+	AnnotationWarning AnnotationSeverity = "warning"
+	AnnotationError   AnnotationSeverity = "error"
+)
+
 type catalog map[string]string
 
 // this catalog is used to translate text from basic terminals to enhanced ones that support emoji, just
@@ -53,6 +62,8 @@ type Output struct {
 	OutputType Type
 	cat        catalog
 	w          io.Writer
+	// severity is the severity used for GitHub annotation commands.
+	severity AnnotationSeverity
 }
 
 // ValidTypes returns an array of the valid output types.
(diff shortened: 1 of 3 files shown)


output/output.go
```diff
@@ -73,18 +84,39 @@ func (o *Output) Printf(format string, a ...interface{}) error {
...
 	_, _ = fmt.Fprintf(o.w, "%s", s)
 	return nil
 }
 
+// githubCommand formats a GitHub Actions workflow command (annotation).
+func githubCommand(severity, message string) string {
+	return "::" + severity + "::" + escapeWorkflow(message)
+}
+
+// escapeWorkflow URL-encodes the characters that are not allowed inside a
```
fzipi, line 108: Why not just using urlencode?

A13501350: The runner only decodes three sequences and renders anything else literally.

fzipi:
@coderabbitai full review

fzipi (approved):

PR #669, opened 2026-09-08
docs: add ADR for benchmark mode

### LINKED ISSUE #143: Allow benchmark mode
Right now go-ftw executes requests one by one and then pull logs expecting the output to so the assertion.

While this is good, sometimes you want to profile and/or measure performance with close to real test data and go-ftw is a good option for this.

I wish go-ftw includes an option where N requests can be sent concurrently. This would allow to collect  profiling data to debug latencies when evaluating and triggering rules.

fzipi:
## what

Adds an ADR proposing a `bench` mode: fire requests from existing YAML test
files concurrently and report aggregate latency/throughput stats, without
per-request WAF log correlation.

## why

Closes #143. Concurrent requests would interleave the start/end log markers
`run` currently uses for per-request correlation (`runner/run.go`), so
correctness checking under concurrency needs a different mechanism (a
per-transaction ID from the WAF connector) that's out of scope here. This ADR
scopes the change down to what the issue actually asked for: profiling and
latency data, not concurrent regression assertions.

## refs

- #143

cc @jcchavezs — would like to iterate on this with you before it's built.

## Summary by CodeRabbit

* **Documentation**
  * Updated the benchmark-mode architecture decision record.
  * Documents concurrency at the individual test-case level rather than the YAML-file level.
  * Specifies that test cases are scheduled up to the configured in-flight limit, while stages within each multi-stage test remain sequential.

### DIFF
--- a/docs/adr/0001-benchmark-mode.md
+++ b/docs/adr/0001-benchmark-mode.md
@@ -34,8 +34,8 @@ The reported need is profiling and latency data, not concurrent regression asser
 ## Decision
 
 Add a benchmark mode that fires the requests from existing YAML test files concurrently, up to a
-configurable concurrency limit, and reports aggregate latency/throughput stats
-(min/max/avg/percentiles, requests/sec) instead of per-test pass/fail.
+configurable concurrency limit, and reports aggregate latency/throughput stats instead of
+per-test pass/fail.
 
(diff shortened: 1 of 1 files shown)


fzipi:
@coderabbitai full_review

theseion (approved):
❤️

